PR #52, opened 2022-10-18
Update +page.svelte

trafficpest:
class tooltip and tooltiptext have been added to css in proposal.

There is tooltips in tailwind if desired requires popper.js

**Screenshots**

**Additional context**


secondl1ght:
We are using a new package now for tooltips so I will add this information using that.

PR #114, opened 2023-11-25
Use preferred tile.openstreetmap.org URL

danieldegroot2:
I would like to suggest using the now preferred https://tile.openstreetmap.org URL instead of the current one (`{s}`), see

https://github.com/openstreetmap/operations/issues/737

### DIFF
--- a/src/lib/map/setup.js
+++ b/src/lib/map/setup.js
@@ -71,7 +71,7 @@ export const toggleMapButtons = () => {
 export const layers = (leaflet, map) => {
 	const theme = detectTheme();
 
-	const osm = leaflet.tileLayer('https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png', {
+	const osm = leaflet.tileLayer('https://tile.openstreetmap.org/{z}/{x}/{y}.png', {
 		noWrap: true,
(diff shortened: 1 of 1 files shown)


secondl1ght:
Hi thanks for the PR! I was not aware of this change so thanks for sharing. There are other URLs in the code that also need to be updated, would you like to add those as well to this PR?

danieldegroot2:
This is the only URL for `tile.openstreetmap.org` [as far as I can see](https://github.com/search?q=repo%3Ateambtcmap%2Fbtcmap.org+.tile.openstreetmap.org&type=code).
(noteably `openstreetmap.org`, not `opentopomap`, `.de` etc., those are different providers)
Otherwise, please provide a link to to the line in the code which needs to be changed.

If there is anything else which needs to be changed (elsewhere this was not the case), then it may be best to check each provider / file manually and change this yourself. This also applies to any other service and attribution.
(note I am not a developer, I just change the URL where possible)

secondl1ght:
Ah ok you are right, I didn't notice in that discussion thread you linked but yes it seems it is only for the `.org` domain. Merging now!

secondl1ght (approved):

PR #128, opened 2024-02-20
feat: link map image

escapedcat:
No

I tend to click on the map image to go to the map on desktop.  
Maybe I'm the only one. Just wanted to propose it.

### DIFF
--- a/src/routes/+page.svelte
+++ b/src/routes/+page.svelte
@@ -76,13 +76,15 @@
 				</h2>
 			</div>
 			{#if typeof window !== 'undefined'}
-				<img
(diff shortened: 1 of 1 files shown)


secondl1ght:
Hey @escapedcat good to hear from you. :) I don't see an issue with this, the only thing might be that we would not want to have this a link on mobile to avoid users accidentally clicking on it while scrolling the page. But I think it's fine, and we can make that update later if anyone complains about this. Thanks for the PR!

escapedcat:
Let me know if this is causing issues. Happy to adjust this.
Thanks for this, really helpful!

PR #131, opened 2024-03-13
Add help links

bubelov:

**Screenshots**

**Additional context**

### DIFF
--- a/src/components/IssueCell.svelte
+++ b/src/components/IssueCell.svelte
@@ -29,4 +29,13 @@
 	>
 		Edit
 	</a>
+{:else if id === 'helpLink' && value}
+	<a
+		href={value}
+		target="_blank"
+		rel="noreferrer"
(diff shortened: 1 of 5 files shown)


secondl1ght:
I was thinking we could just turn the description ex. 'Outdated location' into a link that leads to the Wiki. What do you think? Is it better to have a dedicated column?

bubelov:
I'm not sure if the description field is necessary since we have a type field. It would be stupid to send 100+ descriptions when we'll need to localize the app, and it's not good for sync and local storage

bubelov:
So, the description column can be removed in the future, or it can be derived from the type field client-side. As for the link, I don't really care, but it's better if we put something actionable there as soon as possible

secondl1ght:
This is a good idea about deriving the descriptions from the types on the client-side. Just let me know when you want to deprecate that field in the backend and I will make the updates on the frontend.

Issue #161, opened 2024-10-15
DEV: error after build and start in dev mode


ildella:
**Describe the bug**

I installed all packages and run the build and dev scripts

**To Reproduce**

**Expected behavior**

**Screenshots**
I get a 500 internal error on `dev` and this on the console during `build`

```shell
node:internal/event_target:1090
  process.nextTick(() => { throw err; });
                           ^
ReferenceError [Error]: document is not defined
    at createTag (/home/ildella/projects/contrib/btcmap.org/node_modules/lottie-web/build/player/lottie.js:30:5)
```
**Additional context**

I have an ".env" file but all constant are empty strings.

escapedcat:
Same, looks like [svelte-lottie-player](https://github.com/LottieFiles/svelte-lottie-player/issues) is outdated and unmaintained.

PR #163, opened 2025-01-13
fix: remove lottie to enable build again #161

escapedcat:
#161 

Build was failing due to outdated and unmaintained dep.
TS errors were already ignored before but now it's just breaking.

**Additional context**

If you're ok with leaving out these animations for now we're good. If not a different solution needs to be discussed maybe.


escapedcat:
💁‍♂️ @bubelov

bubelov:
I'm fine with removing animations if it doesn't break anything. Being able to build an app is clearly more important

escapedcat:
To be honest, due to the state of the app I couldn't check how it used to look like.
My assumption is just that animations are usually a nice-to-have and are not crucial for the user.
Hoping you have more insights into this :)

Happy to continue with cleaning up deps and setup.

bubelov:
I tested boosts and they're still working in this branch

PR #164, opened 2025-01-13
chore: non-breaking deps 🤞 

escapedcat:
- Updating minor deps
- Updating further deps

Svelte and Vite updates will be [more effort](https://svelte.dev/docs/svelte/v5-migration-guide#Migration-script).
eslint/prettier could maybe switched to https://biomejs.dev which is faster and easier to maintain. Need to check.


escapedcat:
💁‍♂️  @bubelov

Issue #82, opened 2023-04-15
Fix dark mode map nav icons


dadofsambonzuki:
![Screenshot_from_2023-03-30_11-34-28.png](https://user-images.githubusercontent.com/87125117/232182252-c29ade77-33df-48c7-8fb9-a1eb5884d78a.png)

![Screenshot_from_2023-03-30_11-35-05.png](https://user-images.githubusercontent.com/87125117/232182251-dfbd31f0-95a5-496b-a539-ec034cc94a1a.png)

Pajko97:
@dadofsambonzuki Hi Nathan, as I understand you want the links on images 1 to look like the image 2 with border being same on all sides?

dadofsambonzuki:
The general issue is shadowing not working in dark mode for these assets. I'm not sure shadowing is a design pattern for dark mode and so a solid outline would be better.

vivekmishra9631:
Heyy, I am interestedto fix this bug

secondl1ght:
Hi @vivekmishra9631, sounds good thanks! When you submit the PR it can be reviewed.

dadofsambonzuki:
Hey @vivekmishra9631 - are you still interested in picking this one up?

gillohner:
@dadofsambonzuki I'm looking into fixing this issue right now. 

`.leaflet-bar a:last-child` has `border-bottom: none;` set. 

From my testing it seems like either disabling this or overriding it with something like this should fix the issue:

`borderBottom = '1px solid #ccc';`

I was able to fix the community maps by adding: 
`locateDiv.style.borderBottom` = '1px solid #ccc';`
to the setup.ts

Is this the right way to approach this issue? I'm not very familiar with the codebase :)

dadofsambonzuki:
You're now more familiar than me!

Can you post some screenies or link to a dev env?

Or just submit the PR and we can test on staging.

dadofsambonzuki:
Fixed in https://github.com/teambtcmap/btcmap.org/pull/170

PR #170, opened 2025-01-30
add borders to control-icons to fix shadow

gillohner:
https://github.com/teambtcmap/btcmap.org/issues/82

It makes sure the border: none property is not applied, by setting the border bottom to '1px solid #ccc' on the icons. Currently the borders at the bottom don't look right.

**Screenshots**
image

**Additional context**

### DIFF
--- a/src/lib/map/setup.ts
+++ b/src/lib/map/setup.ts
@@ -404,6 +404,7 @@ export const changeDefaultIcons = (
 		};
 	}
 	fullscreenButton.classList.add('dark:!bg-dark', 'dark:hover:!bg-dark/75', 'dark:border');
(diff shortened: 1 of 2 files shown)


escapedcat:
Thanks @gillohner !  

Does this affect the light-mode somehow?
Can you check if it's possible to achieve this with the used [tailwind classes](https://tailwindcss.com/docs/border-width) like it's done here?:  
https://github.com/teambtcmap/btcmap.org/blob/main/src/lib/map/setup.ts#L272-L277
It might just work with `dark:bottom-b`

escapedcat:
Thanks for your feedback! Just wanted to double check. I'm fairly new to this project.
Let's go with this. We can always adjust it later.

PR #179, opened 2025-02-24
Add vector tile support

bubelov:
Experimental branch, needs review and feedback

Known issues: https://github.com/maplibre/maplibre-gl-leaflet/issues

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,11 +56,8 @@ Thanks to:
 
 ### Base map tiles
 
-- [OpenStreetMap](https://www.openstreetmap.org) (OpenStreetMap, OpenStreetMapDE, OpenStreetMapFR)
-- [OpenTopoMap](https://opentopomap.org) (Topo)
-- [ESRI](https://www.esri.com) (Imagery)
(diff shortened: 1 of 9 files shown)


dadofsambonzuki:
Have tested in a variety or browsers and devices and looks good. Better UX and I actually prefer the new dark tiles.

We need to sync the initial tile selection to the Dark/Light mode setting on the Web App. It currently defaults to a light map when in dark mode.

escapedcat:
Tested on chrome canary. Feels great! Code lgtm

bubelov:
@dadofsambonzuki it defaults to raster tiles, since vector layers are untested. There are some annoying issues with Leaflet vector plugin, like maxZoom being ignored and map starts to jitter once you zoom in too much. I have no idea if it can be fixed but we can probably live with that.

@escapedcat if everyone is okay with that, I'll merge it with the following changes:

1. Vector layers will be default, with an optional fallback to OSM raster. We can use dark vector style for a dark mode and light style for the light mode.
2. Removing Stadia copyright

PR #188, opened 2025-02-28
Feat/add location address optional

### LINKED ISSUE #187: Remove requirement for address on Add Location form
Lots of places don;t actually have an address.

Lat/Lon should be required.

escapedcat:
Fixes #187 

- noticed that location select is already required
- connected some labels to their inputs

### DIFF
--- a/src/routes/add-location/+page.svelte
+++ b/src/routes/add-location/+page.svelte
@@ -408,18 +408,13 @@
 							</div>
 
 							<div>
-								<label for="address" class="mb-2 block font-semibold"
-									>Address <InfoTooltip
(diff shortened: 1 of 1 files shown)


dadofsambonzuki:
Maybe move address down below location and put option in brackets to make it consistent?

We should remove twitter as a field and change to prompt in the notes field to add any social media there.

Cc @stackingsaunter

dadofsambonzuki:
Covered this in my current refactor of the forms pages as part of the switch to Gitea for data Issues.

Have ensured the input elements have IDs too as per your change.

PR #191, opened 2025-03-12
Sorting

GuiMF:
Created a button to sort countries and communities per capita

### DIFF
--- a/src/components/AreaLeaderboard.svelte
+++ b/src/components/AreaLeaderboard.svelte
@@ -2,7 +2,7 @@
 	import { AreaLeaderboardItem, AreaLeaderboardSkeleton, GradeTable, TopButton } from '$lib/comp';
 	import { areaError, areas, reportError, reports, syncStatus } from '$lib/store';
 	import type { Area, AreaType, LeaderboardArea, Report } from '$lib/types';
-	import { errToast, getGrade, validateContinents } from '$lib/utils';
+	import { detectSort, errToast, getGrade, validateContinents } from '$lib/utils';
 	import tippy from 'tippy.js';
 
 	export let type: AreaType;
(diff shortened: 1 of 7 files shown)


escapedcat:
Thanks!  
Also thanks for you comments inside your code to explain your reasoning. You could switch those from the code to the PR comments to avoid letting them get merged.

- Have you tried to optimize your code asking AI for help? I'm also new to svelte and it helps to ask i.e. "This is how I do stuff in react, how would I do it in svelte?"
- Not sure if a whole extra button is needed. Can you try to avoid it? Looks like you copied an existing component to add you're functionality? This should be separated if possible.
- Would you mind adding screenshots of your feature to the PR?
- With CSS we can help, no worries.

GuiMF:
The Button I created was unnecessary, I just used the primary button and added a function to the click. I don't know if the way I imported the function was right but it works, I tried to add it to the utils but it wasn't working... Also removed comments... I'll add screenshots once I have better wifi, the stats seem to not be loading.

GuiMF:
Screenshot 2025-03-12 at 17 19 33 This is what sorting by GDP looks like. Pressing the "Toggle Sort" toggles it back to sorting by total locations. It still displays the locations by GDP.

dadofsambonzuki:
I think we can just remove the button and have sort up/down arrows on each relevant column.

People often want to order by absolute number too.

Position column should still represent the default sort, which is a weighted number based on up-to-datedness.

We also need to handle the case where there is no population tag present.

We should also apply this to the country pages.

GuiMF:
That does sound like a good idea to make the UI simpler, I used a button because I didn't know how to create the arrows.

As for the grade problem that's a bit of an issue because the grade is simply given to the top 3 objects in the list, with that said the grade would have to be a property given to the object itself and not the position of the leader board. 

The no population tag issue is something I completely forgot about... I'll take a look!

This has already been implemented on the country pages.

bubelov:
Yeah, all the data is cached locally so we can re-arrange it as we like, but we're slowly moving to a view-based API where the website loses offline mode/caching but gains the ability to fetch everything needed to populate the UI on the fly. That new API can support pagination easily

escapedcat:
Even more reasons to try out the tanstack-table approach 👍

escapedcat:
Continued here: #272

PR #202, opened 2025-04-06
chore: add tsc task and split up lint tasks

escapedcat:
- also add tsc task to GH actions ~but let it fail for now~
- enable the existing svelte-check task in ci (before it was never really run)
- ignore or fix current TS issues
- let future CI runs fail if a type issues comes up

image


bubelov (commented):
Looks good to me

PR #222, opened 2025-04-16
Style/lint svelte each key after revert

escapedcat:
Need to test this in more detail


escapedcat:
Will get back to this at a later stage

Issue #229, opened 2025-04-27
Show comments on merchant pages


dadofsambonzuki:

bubelov:
API is ready:

https://api.btcmap.org/v4/places/way:1345144584/comments

https://api.btcmap.org/v4/places/22923/comments

Both new and old (OSM) place IDs are supported

PR #239, opened 2025-05-10
feat(merchant): display comments #229

escapedcat:
Resolves: #229

- copied `MerchantEvent` and adjusted it
- switch from client API call to server component
- added type for the pageload-data

Notes:
Not exactly sure about the design. I searched some merchants with more or longer comments but couldn't easily find one. Can adjust if needed/wanted.

image

https://deploy-preview-239--btcmap.netlify.app/merchant/node:3708729816

### DIFF
--- a/src/components/MerchantComment.svelte
+++ b/src/components/MerchantComment.svelte
@@ -1,8 +1,9 @@
 <script lang="ts">
 	import Time from 'svelte-time';
+	import type { MerchantPageData } from '$lib/types';
 
-	export let text: string;
-	export let time: string;
+	export let text: MerchantPageData['comments'][number]['text'];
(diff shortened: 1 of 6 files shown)


dadofsambonzuki (commented):
Looks good.

I'd move the comments section above the activity section.

dadofsambonzuki:
I see the MerchantData interface which is a good stepping stone to moving away from Element to a more BTC Map centric interface for those pages.

dadofsambonzuki (approved):

escapedcat:
> I see the MerchantData interface which is a good stepping stone to moving away from Element to a more BTC Map centric interface for those pages.

That's actually just to have a type for the incoming data there which we hadn't before. But we might be able to adjust/extend this later one maybe if we switch to v4. 🤞

Issue #244, opened 2025-05-13
Add 'Add comment' feature to merchant page.


dadofsambonzuki:

escapedcat:
> https://github.com/teambtcmap/btcmap-api/tree/master/docs/rest-api#paywall_add_element_comment

This link doesn't seem to work for me. It opens the overview.

I found [the RPC doc](https://github.com/teambtcmap/btcmap-api/blob/master/docs/rpc-api/element-methods.md#paywall_add_element_comment) but I don't know the URL.
Looking at i.e. [the current boost code](https://github.com/teambtcmap/btcmap.org/blob/main/src/components/Boost.svelte#L90-L92) I'm not sure how to map the URL to the RPC/rust function.

dadofsambonzuki:
Sorry for delay.

Boosts in the Web App use LNBits right now (Android already done), but we will be migrating to the RPC approach. Add Comment should use the RPC you found. 

URL is `https://api.btcmap.org/rpc`.

PR #249, opened 2025-05-18
feat(merchant): add comment #244

escapedcat:
Fixes: #244

### Follow up

- adjust design of contact and opening hours: #253
- see mentioned refactorings below
- do we have some sort of alert/warning component already (simple icon/text/color-variant)? might create one

#### Initial button
image

#### Modal opened
image

Open points:
- What do you think about the position of the button?
- Regarding modal content I stuck with what is being displayed in the android app
- Lot's of opportunities for refactoring i.e.
    - general button design
    - modal component


dadofsambonzuki:
Nice!

>     * What do you think about the position of the button?

I think we should move it up to the top with the other buttons. I initially thought something like this:

Screenshot from 2025-05-22 21-51-45

But then I realised the top of that page has been bothering me for a while. I think we should make the opening hours less prominant and more like the accepted payments. That would then lead the way for Boosts, Last ~Surveyed~ Verified and Comment (perhaps showing the comment count?) to be displayed as an action trio across the page.

>     * Regarding modal content I stuck with what is being displayed in the android app

We should align it with the Boost copy, and so 'Comment' rather than 'Generate Invoice'. We should also change to button text when pressed like the Boost.

escapedcat:
Current state:

image

I guess we could move contact and opening hours next to accepted payments?
Not sure if opening hours can be longer than one line though. Could look awkward if there are several lines.

image

escapedcat:
image

bubelov:
Just tried it, worked like a charm

escapedcat:
@dadofsambonzuki we good here?

dadofsambonzuki:
> @dadofsambonzuki we good here?

Grrr, a replied with some feedback last night but it didn't post for some reason.

TLDR - Good to go. UX can be improved with payment success feedback, but the only way we can do that currently is by polling the LNBits server directly like this:

```GET https://core.btcmap.org/api/v1/payments/<payment_hash> -H "X-Api-Key: ac9503a3628c4e21abc8708bb36f0ec0" -H "Content-type: application/json"```

Perhaps once a second for five seconds then either confetti on success or "Payment hasn't yet been received, we'll continue checking and post your comment when paid".

Issue #235, opened 2025-05-08
eslint warnings


dadofsambonzuki:
There are hundreds of eslint warnings throughout the project.

A thankless task, but some hero should address them.

escapedcat:
Status update (after #240 is merged):

```sh
✖ 213 problems (0 errors, 213 warnings)
```

These are coming from the left over rules.
A good start would be to tackle the rule `'svelte/require-each-key': 'warn',`

Steps:
- comment rule out [here](https://github.com/teambtcmap/btcmap.org/blob/main/eslint.config.js#L35)
- run `yarn lint`
- fix the svelte loops one by one while testing if each view/component is still working as expected

Let us know if you need any help or you have any questions regarding this task

kwilson62:
Anyone working this yet? If not, I can start working through them as I learn the code base.

escapedcat:
@kwilson62 go ahead, let us know if you get stuck

escapedcat:
Update after #257 

> Before:
> 
>     ✖ 194 problems (0 errors, 194 warnings)
> 
> After:
> 
>     ✖ 127 problems (0 errors, 127 warnings)

escapedcat:
[Example](https://github.com/teambtcmap/btcmap.org/commit/62a1f35134bbd21eeeea0bc6cf9724bd1568c159) on how to fix (some?) `@typescript-eslint/no-unused-expressions` warnings.

Via: https://github.com/sveltejs/eslint-plugin-svelte/issues/934#issuecomment-2543128219

escapedcat:
[Further fixing examples](https://github.com/teambtcmap/btcmap.org/commit/5c89931e4e0df905cfc3440b6653d61c713adb6b)

(AI helped)

PR #257, opened 2025-05-28
Chore: Resolve svelte/require-each-key Warnings

kwilson62:
Related: https://github.com/teambtcmap/btcmap.org/issues/235

Resolved all `svelte/require-each-key` warnings

Note: I did check each component and page that was edited, however I was unable to verify some components like the table on the "Open Tickets" page due to not having the Gitea env variables. That said, the skeleton still loads and looks as expected.

These changes also passed `yarn format` and `yarn playwright test`

But I would appreciate if someone could check out these changes to make sure components / pages that require .env variables render correctly.

Before:

> ✖ 194 problems (0 errors, 194 warnings)

After:

> ✖ 127 problems (0 errors, 127 warnings)

### DIFF
--- a/src/routes/about-us/+page.svelte
+++ b/src/routes/about-us/+page.svelte
@@ -269,7 +269,7 @@
 
 				<div class="flex grid-cols-3 flex-wrap justify-center gap-5 lg:grid">
 					{#if merchants.length}
-						{#each merchants as merchant}
+						{#each merchants as merchant (merchant.id)}
 							<AboutMerchant
 								id={merchant.id}
 								icon={merchant.tags['icon:android']}
@@ -278,7 +278,7 @@
 						{/each}
 					{:else}
 						<!-- eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars -->
-						{#each Array(6) as skeleton}
+						{#each Array(6) as _, index (index)}
 							<span class="h-24 w-24 animate-pulse rounded-full bg-link/50" />
 						{/each}
 					{/if}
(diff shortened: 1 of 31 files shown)


src/routes/about-us/+page.svelte
```diff
@@ -278,7 +278,7 @@
 						{/each}
 					{:else}
 						<!-- eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars -->
```
escapedcat, line 280: Can this be removed as well?

src/routes/about-us/+page.svelte
```diff
@@ -338,12 +338,12 @@
 				<div>
 					<div class="flex grid-cols-3 flex-wrap justify-center gap-5 lg:grid">
 						{#if communities.length}
-							{#each communities as community}
+							{#each communities as community (community.id)}
 								<AboutCommunity {community} />
 							{/each}
 						{:else}
 							<!-- eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars -->
```
escapedcat, line 345: Can this be removed as well? (Maybe have a look if you find more occurrences which can be removed as well)

escapedcat (commented):
Awesome, thanks!

Looks like we can now [remove the rule](https://github.com/teambtcmap/btcmap.org/blob/main/eslint.config.js#L34) now to make this an error again in the future

escapedcat:
> Note: I did check each component and page that was edited, however I was unable to verify some components like the table on the "Open Tickets" page due to not having the Gitea env variables. That said, the skeleton still loads and looks as expected.

Looking good 🙌 

image

escapedcat:
Thanks!

PR #266, opened 2025-06-01
feat(merchant): aligh opening hours

escapedcat:
### Before
image

### After
image

### Further ideas

Adjust it here too?  
image


dadofsambonzuki:
After looks marginally better, but I think the icon here and next to contact are generally misplaced. Perhaps would look better as part of the title, which would also give more control over centering of the actual content?

Also, what do you think to this to break up the page a little and put key information at the top?

Screenshot from 2025-06-01 06-27-57

The action trio could become cards too?

escapedcat:
> The action trio could become cards too?

Could be done.

You mean such a card? 
image
(i.e. "Comment" is headline. Text and Button in the card-body)

Or just white round borders?

dadofsambonzuki:
> Or just white round borders?

Yeah, just this I think. Line break as per communities cards a little too much here.

dadofsambonzuki:
Yes, much better. Does that mean we can centre opening house content now?

PR #272, opened 2025-06-06
Feat/leaderboard tenstack table

escapedcat:
https://deploy-preview-272--btcmap.netlify.app/communities/leaderboard

Notes:
- [x] mobile view improvements

## Draft 5
image

image

image

## Draft 3

image

## Draft 2 (with styling from [tagging-issues](https://btcmap.org/tagging-issues))

image

## Draft 1

image


dadofsambonzuki:
Looks good.

1) We can keep the algo description at the bottom as it's clear with the new styling and pagination.

2) Lets not loose the tooltips on the column headers.

3) Looks like Grade is secondary sorted by total locations, this should be by up-to-date percentage.

4) We can loose the up-to-date percentage column and have that as a tool tip on grade.

5) We should add in a `Up-to-date Locations` column.

6) Description of the algo should be:

```
Position is calculated as follows:

    Primary: Total locations minus 5x outdated elements.
    Secondary: Total number of locations.

Locations include ATMs.
```

The changes to the Country Leaderboard will be identical and are handled in the same component. Shall we extend the scope here or as a seperate PR?

escapedcat:
- > * Looks like Grade is secondary sorted by total locations, this should be by up-to-date percentage.
       
   How do you see this? Looks like it's just done [via `up_to_date_percent` ](https://github.com/teambtcmap/btcmap.org/blob/feat/leaderboard-tenstack-table/src/components/AreaLeaderboardMinimal.svelte#L103)
-  > * 4) We can loose the up-to-date percentage column and have that as a tool tip on grade.
    
     Like this right? ✅ 
      image
- > The changes to the Country Leaderboard will be identical and are handled in the same component. Shall we extend the scope here or as a seperate PR?
   
    Will do this in a follow-up PR 


For other points see latest draft above in the desc

dadofsambonzuki:
> * How do you see this? Looks like it's just done [via `up_to_date_percent`](https://github.com/teambtcmap/btcmap.org/blob/feat/leaderboard-tenstack-table/src/components/AreaLeaderboardMinimal.svelte#L103)

Yes, that's how the Grade is calculated. But once sorted by Grade, it seems like its _secondary sorting_ on total_locations. This should be up_to_date_percent as the finer grade version of Grade.

dadofsambonzuki:
> * Like this right? ✅

The tooltip on the column header is great for the legend. I'm talking about hovering over the actual Grade value for a given row and getting the up_to_date_percentage.

dadofsambonzuki:
A couple of other points:

7) Maybe the algo description should be a tooltip on the Position column header?

8) Location headers should be `Total Locations`  (Tooltip: All locations inc. ATMS) and `Verified Locations` (Tooltip: Locations verified within the past year).

9) Include an `Average Verification Age` column to the left of Grade, which is `today - average_verification_date`.

escapedcat:
- >  Maybe the algo description should be a tooltip on the Position column header?
    
     Hm, I feel like that's a lot of text for the current tooltip. Currently I'm not sure how to make those work on mobile, so crucial information might be missing on mobile.
-  > > Like this right? ✅
       > 
       > The tooltip on the column header is great for the legend. I'm talking about hovering over the actual Grade value for a given row and getting the up_to_date_percentage.
            
    image
-  > Include an Average Verification Age [...]
      
     Added but now with this plus the "Location"-column header adjustments the table gets very wide now. Need to think about how and what to display on mobile.

dadofsambonzuki:
We could move "Average Verification Age" to the Grade tooltip as it is a quality metric. Tooltip would show % and age.

Moving also discourages communities needlessly re-verifying locations too frequently to get that number as low as possible if it is not as prominent.

escapedcat:
Are we good design//feature wise? Then I'll have a look at the general code and switch this to a PR.

escapedcat:
image

dadofsambonzuki:
Hmmm, definetely better, but I'm still seeing community names rolling into the next column.



We could combine Total/Verified into a single column I guess so that we have like 42/21.

We could also maybe put the name under the icon?

Would be good to tweak the table rather than have to maintain a completely different design.

No sorting on mobile either.

dadofsambonzuki:
Or auto landscape, but that is probably some design no no.

escapedcat:
wdyt?

image

dadofsambonzuki:
That could work if we centered the logo/name row? Or perhaps better to have the logo on its own row (1st) so it can be bigger?

That would enable total verified to be split and enable separate sorting?

Issue #276, opened 2025-06-09
Improve map cold start time and make it reliable


bubelov:
I keep getting complaints about users unable to check BTC Map because it simply fails to load. It was also reported by Wallet of Satoshi, which is embedding BTC Map and their users are reporting the same issue.

I don't really know what's the root cause, but in some cases modal get stuck, so we can simply remove it, I guess? I mean, people report seeing places, but they can't interact with the UI due to that modal which can't be closed or cancelled.

I just fired up my browser in anon tab and opened this link:

https://btcmap.org/map

15 MB transferred, 15 seconds to load on a beefy PC with 1 GBit link.

Most of those 15 MB are sync (v2 place sync taking 5 MB)

I think it would be better to make sync fully (A)sync without any UI blocking.

Also, it's possible to cut map loading time dramatically by using v4:

https://github.com/teambtcmap/btcmap-api/blob/master/docs/rest/v4/places.md

```bash
curl --request GET https://api.btcmap.org/v4/places?fields=id,lat,lon,icon,comments
```

The full output contains all ~16,000 active merchants and it has everything to draw and cluster the pins. It takes about 200 kB compressed, that would be a huge performance gain and it's also much easier to parse:

```json
[
  {
    "id": 9918,
    "lat": 43.0332123,
    "lon": -88.1729769,
    "icon": "computer"
  },
  {
    "id": 3430,
    "lat": 10.3406452,
    "lon": 123.9351582,
    "icon": "lunch_dining"
  },
  {
    "id": 3431,
    "lat": 10.3406879,
    "lon": 123.9351169,
    "icon": "lunch_dining"
  },
  {
    "id": 3432,
    "lat": 10.3407271,
    "lon": 123.9350893,
    "icon": "lunch_dining"
  },
  {
    "id": 3442,
    "lat": 14.7264989,
    "lon": -91.2631044,
    "icon": "local_cafe"
  }
]
```

Fields that aren't present are omitted, for efficiency.

bubelov:
Given the size of the data, v4 approach can give us few more years, I think it makes sense to try to load everything while we can, but at some point we will be forced to limit min zoom and only load small regions within user's view port, probably after passing 100-500k merchants

bubelov:
> So it looks like currently we get "all ze data" for each element (i.e. opening hours) which is needed to show the content of the pin-popup on the map:

[...]

PR #307, opened 2025-07-08
Feat/276 improve cold start

escapedcat:
Eventually might fix: #276

Test URL: https://deploy-preview-307--btcmap.netlify.app/map?13/52.53398/13.41602#14/52.55350/13.40409

Notes:
- search on map is gone
- verify direct link is gone
- other views than map which depended on elements-store use v2/v3 API call for now
- popup 
    - more link in  is gone
    - extend boost icon has been changed by claude

### Screenshots

| Prod      | PR      |
| ------------- | ------------- |
| image | image |
| image | image |


bubelov:
You can easily request any OSM tag via `osm:` prefix. If you need `payment:*` tags, you can add those fields like that: `/places/fields=osm:payment:ligntning`. Those tags are unmaintained and unreliable though, so I don't recommend requesting them, but the API doesn't filter secondary tags, you just need to specify the data source. Places aren't elements, we will probably add meetups, conferences, events, etc, so it makes no sense to treat merchant-specific tags as something you should expect to get from any kind of place

escapedcat:
> I still see v2 API being called, maybe it triggers that blocking popup

Wanted to start with getting rid of elements first. All others (i.e. events) are still on v2

> I think it's fine to disable certain features and even pages for some time, we can review them one by one and decide if we even need them. 

I'm ok with that

> @dadofsambonzuki mentioned that the alternative approach would be to re-create the web app from scratch, so it would be even more limited initially.

Yes, I thought it would be nicer to first update the current version a bit and use that as a base to refer to for a redesign.

> Those tags are unmaintained and unreliable though, so I don't recommend requesting them, [...]

This seems to be an ongoing discussion :D  
Good with either tbh. Stuck between being clean and stable (get rid of stuff) and not confuse users with removing things they are used (and maybe rely on even you should not?) 🤷‍♂️

escapedcat:
Re-added the missing info to the popup for now to keep things as is. Once we have a decision how to move on we can adjust this.

Some feedback from claude after updating some missing fields. Just fyi. I'm not sure if we still need or want any of those:
```sh
  Missing OSM Data Fields:

  - description - Place description text
  - note - Additional notes about the place
  - osm:payment:lightning:requires_companion_app - Third-party app requirement flag
  - osm:payment:lightning:companion_app_url - URL for required companion app
  - issues - Data quality issues array
  - areas - Area/region associations

  Missing Verification Fields:

  - survey:date - Manual survey date
  - check_date - General check date
  - check_date:currency:XBT - Bitcoin-specific check date

  Missing Category/Classification:

  - category - Merchant category (restaurant, shop, etc.)
  - Full amenity classification might be limited

  Missing Payment Method Details:

  - payment:lightning:requires_companion_app - Third-party app requirement
  - payment:lightning:companion_app_url - Companion app URL

  Missing Metadata:

  - boost:expires - Boost expiration (we have boosted_until instead)
  - Full OSM bounds data for ways/relations
  - Complete OSM tag set for advanced filtering

  ✅ Already Available in v4:

  - address - Replaces structured addr:* components
  - verified_at - Replaces survey/check dates
  - boosted_until - Replaces boost:expires
  - All basic contact fields (phone, website, email, social media)

  So the structured address parsing is actually covered by the single address field in v4, which is probably more practical anyway.
```

escapedcat:
> `boost:expires` is the same thing as `boosted_until`
> verification fields are merged into a single field called `verified_at`

Yeah, these are fine


> `required_app_url` replaces all the companion app fields

Got it. As far as I understand we currently do not show the layer. [The  #functionality for that is commented out according to this issue](https://github.com/teambtcmap/btcmap.org/issues/308).

 
> where do you need issues and areas? Map page or a place page? Do you need counts or full structs?

I checked again.  
- areas: For the map we do not need the areas, so we're good for now.
- issues: Issues are currently used like this I think:
   image
   image
    https://btcmap.org/map#19/32.33804/-117.05547

bubelov:
I think it's okay do disable or remove that issues button. It's useless in its current form:

1. An average user doesn't know anything about tags or issues, it should be renamed to "data quality issues" or something
2. Those things should be highlighted in red, otherwise they won't be perceived as warnings or actionable items
3. All listed issues should be actionable, there should be a clear set of instrucions on how to fix them
4. We have more convenient issue management UI for power users: https://btcmap.org/country/de (maintain tab, deep links would be cool though)

dadofsambonzuki:
> `required_app_url` replaces all the companion app fields

Do we not need `requires_companion_app` to enable pin filtering?

bubelov:
> Do we not need `requires_companion_app` to enable pin filtering?

Isn't `required_app_url` enough? It implies a third-party app is needed

escapedcat:
> Yep, it's linked to from merchant pages of clients with pre-filled element info.

Yeah, that's cool and makes sense. But does the direct link in the navigation make sense? Is anyone using this to select a location?

image

escapedcat:
> See the analytics channel on Discord. 2nd most popular route, but I'm not sure if that includes the direct referrals or not. Presumably they'd show up with query parameters.

Right. Are we willing to get rid of the map-selection root link (Can re-add later if people complain) to reduce this PR?

dadofsambonzuki:
> Right. Are we willing to get rid of the map-selection root link (Can re-add later if people complain) to reduce this PR?

Yeah, i think we can - especially if we want to move to a map-first approach as per the Bitcoin Design convos. Users should just go find the merchant on the map and then do the things. The verify route should arguably only be called with a Place/Element ID as an argument.

dadofsambonzuki:
- [ ] If a place has a comment, the `Add Comment` button should be `View Comments` and link to the comments section.

dadofsambonzuki:
>> This? Or which one?

Nope, on the socials section on the merchant page.

https://deploy-preview-306--btcmap.netlify.app/merchant/node:10981284890

escapedcat:
> Add Comment button down with comments?

Something like this? https://github.com/teambtcmap/btcmap.org/pull/249

escapedcat:
> Nope, on the socials section on the merchant page.

Fixed:
image

The dropdown menu from the marker will be gone anyways. This PRs env is: https://deploy-preview-307--btcmap.netlify.app/merchant/node:11603641996

dadofsambonzuki:
Conditional button perfect 👨‍🍳🤌

escapedcat:
Are we good here? o_0

escapedcat:
@bubelov I also get the CORS issue on the deployed test env:  
image

During local development this worked for some time and then stopped. Any ideas? I would expect this to work at least in the deployed env here

bubelov:
I'm not sure about CORS stuff, but search RPC needs authorization, and it's news to me that the website uses it. Our RPC interface is lower level and admin focused, and it can change with no notice. If website needs something extra, it's better to discuss it in advance and expose it via REST API

I made this method public as a temp fix, can you check if it still stuck?

bubelov:
It could also be related to the `limit` arg, you can try removing it since server knows nothing about it

bubelov:
Getting rid of search would be ideal for now, if above mentioned solutions don't work.

There is also REST search endpoint that was submitted by Blink, I think:

https://github.com/teambtcmap/btcmap-api/pull/47

The problem is, no one even tried to discuss it so it's hard to make it permanent with proper backward compat guarantees since no one wants to discuss it and reach an agreement no how do we want this search API to look like. I don't really know why people try to sneak stuff or hack around instead or just opening a dedicated issue and having a proper discussion =)

bubelov:
Are there any other blockers? I tried preview link and it looks like all the core features are there and the cold start perf is good

We'll probably need a list of removed features, sorted by prio, so we can go trough them one by one and re-enable once this branch is live

escapedcat:
@dadofsambonzuki thoughts?

dadofsambonzuki:
Not had time to look. It surely can't be worse than current perf so I'm happy to merge and then re-add the missing features in smaller PRs.

bubelov:
It doesn't look that the website is using CDN when fetching snapshot though:

Screenshot From 2025-09-06 10-18-01


This request should be taking under 100 ms, not 2+ seconds

Here is what I get with https://cdn.static.btcmap.org/api/v4/places.json

Screenshot From 2025-09-06 10-20-50

escapedcat:
@bubelov in a private tab (clean cache) I see this:

image

(filtered network tab by "plac")

- gets the json
- gets it again (tiny and fast not sure why will check, not blocking)
- gets the rest of the data via v4

Can you check your cache and/or provide a full screenshot of your network tab?

escapedcat:
> Although, that said I think we still have an issue with the tagger page not showing pins.

You're right. Had a look.

Issue seems to be that events only know the `osm_id`, which we do not have in the stored minimal places data.
So we either get the missing data via elements v2 or we find different way.
i.e. 
- make places filterable tagger-id or by `osm_id`?
- ...

I think [events in v4](https://github.com/teambtcmap/btcmap-api/blob/master/docs/rest/v4/events.md) are completely different to [v2 events](https://github.com/teambtcmap/btcmap-api/blob/master/docs/rest/v2/events.md) (basically meetups vs tagging-events). Not sure what the plan is for tagging events.

bubelov:
Tagger map is not necessary and it just duplicates OSM user activity page, which we can reference to if we absolutely must.

And places snapshot is fetched without cdn (`static.btcmap.org` instead of `cdn.static.btcmap.org`) as far as I understand:

Screenshot From 2025-09-06 17-28-20

dadofsambonzuki:
Yep, happy to drop the tagger map for now and re-add later. Tagger pages are cool imo and the taggers themselves like them with badges, tip buttons etc.

bubelov:
Now we're talking :rocket: 

Screenshot From 2025-09-08 09-53-14

Places are now loaded **way** faster than vector tiles, I guess it will be our next bottleneck to fix

So, let's merge it today, unless anyone objects?

escapedcat:
What a 🎢

PR #317, opened 2025-07-16
Community rework sveltish

escapedcat:
- new urls will use i.e. `/countries/europe` which is the right way 
- ~do we need to keep supporting old urls with a `#`? currently this code does, but could get rid of it~
    > Add redirects for backward compatibility

reminder for me:
- ~x logo on 500 page is broken~


dadofsambonzuki:
> do we need to keep supporting old urls with a #? currently this code does, but could get rid of it

No, they didn't even work beyond filtering.

dadofsambonzuki:
Fyi, the tabbed sections on the country and community pages also use the hash approach and need Sveltifying.

PR #367, opened 2025-09-02
feat(map): re-add search functionality with API-based backend #276

escapedcat:
After [this discussion](https://github.com/teambtcmap/btcmap.org/pull/307#issuecomment-3244386094) [the previous PR](https://github.com/teambtcmap/btcmap.org/pull/340) was reverted

~Searching for "pizza" and i.e. choosing "Pizza 911" leads to an area without a place. Is the search API data outdated? Or are the places not in sync somehow? Most places work like expected.~


bubelov:
The discussion above mentioning an empty issue: https://github.com/teambtcmap/btcmap-api/issues/46

This issue was silently added and silently closed with no spec and no public discussion, I have no idea what that endpoint does. Maybe where were some private discussions I'm unaware of. I suggest re-opening this issue and writing proper spec, which should include at least:

- use cases
- arguments
- output entity types (places, areas, etc) and required fields

Then we can try to figure out how to implement a good search endpoint which can be used across all platforms

bubelov:
FYI: I just finished places search endpoint we promised to one of our sponsors, it's better to use it for now since it's the only REST search endpoint which was specified and discussed (it's not on GH unfortunately).

https://github.com/teambtcmap/btcmap-api/blob/master/docs/rest/v4/places.md#search-places

This endpoint is strictly places-only, and it will also return pending places in the future. Pending places are too dynamic to be included in time-based sync endpoint and CDN snapshots, but since search output is cheap and ephemeral, we can include more stuff without breaking backward compatibility guarantees for full caching clients.

If global search is desired (areas, events, etc), it needs to be specified first. Reusing RPC endpoints is dangerous since it's a low level API for admins and it can return private data as well as deleted merchants, etc. Some things are not expected to be shown to regular users.

The only unspecified thing about places-scoped search is output format. It currently returns only the most basic fields, but we don't have the same performance concerns for scoped queries and I think it won't hurt to include all or at least most fields by default. Payload size is also not a concern when we're talking ~100 places (avg map viewport) instead of 17,000 places and growing.

PR #402, opened 2025-09-28
perf(map): remove search result fly animation for better mobile UX

escapedcat:
Replace map.flyTo() with map.setView() in searchSelect function to eliminate slow/weird animation on mobile devices. Instant navigation provides better responsiveness and user experience.


bubelov (approved):
Works fine on my devices, the code looks fine too

Issue #378, opened 2025-09-10
Migrate comments to REST API


bubelov:
Our existing comment flow uses RPC which we should eventually move away from, but it also checks invoice status by circumventing API server and acessing LNBits instance, breaking an important abstraction and forcing us to distribute payment node keys among all the devs willing to contribute to that repo, which doesn't scale well (not all devs are trusted initially and we might move away from LNBits or add a backup instance to improve service uptime).

The workflow is very similar, the only thing needs changing is URL and some arg names:

https://github.com/teambtcmap/btcmap-api/blob/master/docs/rest/v4/place-comments.md

The resulting invoice can be looked up by ID via this endpoint:

https://github.com/teambtcmap/btcmap-api/blob/master/docs/rest/v4/invoices.md#get-by-id

So, we can replace all the LNBits hacks with simple polling by id until status changes to "paid", which would mean we received the payment and we can complete the payment UI flow.

PR #409, opened 2025-10-03
refactor(merchant): comment create use rest api #378

escapedcat:
Fixes: #378

### DIFF
--- a/src/components/CommentAdd.svelte
+++ b/src/components/CommentAdd.svelte
@@ -8,6 +8,7 @@
 	import { tick } from 'svelte';
 	import OutClick from 'svelte-outclick';
 	import { fly } from 'svelte/transition';
+	import { invalidateAll } from '$app/navigation';
 	import type { MerchantPageData } from '$lib/types.js';
 
 	export let open: boolean = false;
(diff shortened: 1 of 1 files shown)


dadofsambonzuki:
Nice!

Feels like we should handle the payment receipt confirmation in the modal with the invoice rather than as a toast message. It all happen v quickly! Maybe leave the modal open until user closes and drop the toast pop-up?

escapedcat:
Updated

bubelov:
Looks good, a few notes:

1. Do you have LNBits access in order to issue refunds? If not, we should create or share existing credentials. Comments are noise but a good boost debug session get lead to bankruptcy =)

2. Can we change "Your comment will be published shortly" to page force refresh (or a force refresh hook on dialog close)? If your invoice is detected as paid it means all the server-side hooks were already processed and the comment is live. In other words, hitting F5 will always show that newly posted comment, so we can auto display it instantly. Can be a part of a future improvement ofc since it's not critical

escapedcat:
>     1. Do you have LNBits access in order to issue refunds? If not, we should create or share existing credentials. Comments are noise but a good boost debug session get lead to bankruptcy =)

Nope, paying to test :) (Please delete the TEST comments  :P)

>     2. Can we change "Your comment will be published shortly" to page force refresh (or a force refresh hook on dialog close)? If your invoice is detected as paid it means all the server-side hooks were already processed and the comment is live. In other words, hitting F5 will always show that newly posted comment, so we can auto display it instantly. Can be a part of a future improvement ofc since it's not critical

✅

Issue #377, opened 2025-09-10
Migrate boosts to REST API


bubelov:
Currently, our website has direct access to both low level RPC API as well as to our LNBits instance.

This is obviously very bad, and it makes it impossible for new contributors to even test some features since we can't send our sensitive keys to randos, and even if you're a trusted dev, there is still friction in setting up .env stuff.

REST API is now public and well documented. The workflow is very simple and does not require the posession of any keys, which makes it easy to onboard new devs and also allows us to integrate boosts and other paid features into mobile apps where embedding the keys was never an option, for obvious reasons.

As a cherry on top, we wouldn't need to share our keys and wallet with Netlify 😄 

https://github.com/teambtcmap/btcmap-api/blob/master/docs/rest/v4/place-boosts.md

dadofsambonzuki:
Invoice status can be polled here for UX feedback.

https://github.com/teambtcmap/btcmap-api/blob/master/docs%2Frest%2Fv4%2Finvoices.md

PR #417, opened 2025-10-06
refactor(boosts): boosts rest api #377

escapedcat:
Fixes: #377

- updated to API
- switched FE API routes to use `api` route to make it clear
- made code between comments and boosts more DRY
- merchant view is not updating immediately after paid invoice because of local-storage data update issues
    - this is the case currently anyways
    - will be tackled when we switch merchant views places v4 api

### DIFF
--- a/src/components/Boost.svelte
+++ b/src/components/Boost.svelte
@@ -34,6 +34,12 @@
 		$resetBoost = $resetBoost + 1;
 	};
 
+	const handleOutClick = () => {
+		// Never close the boost modal on outside clicks to prevent accidental loss of progress
+		// Users must explicitly use the close button or complete the boost flow
+		// This protects against accidental clicks during any stage of the boost process
+	};
+
 	let invoice = '';
 	let invoiceId = '';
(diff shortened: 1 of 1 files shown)


dadofsambonzuki:
Niiiice.

A few points:

1) If a user clicks on the map (easy to do on mobile), the invoice modal disappears. If the invoice is paid, I get confetti and this error. We should only allow modal close on x. Probably the same with comments. If modal is consciously closed AND invoice is paid, we should not show the error. 



2) Can we update the status of the pin on success as some people boost from the map. Noting merchant page reactivity will come later with v4.

3) Where are we getting the fiat rates from?

dadofsambonzuki:
1 is reproducible for me on various browsers and devices.

I think if we ensure the modal can only be closed by clicking X we guard against this.

[Screencast from 2025-10-17 16-30-56.webm](https://github.com/user-attachments/assets/1d7c1242-3cb3-4021-93f4-e19c7757d367)

The other points can be follow-ups for sure.

Issue #181, opened 2025-02-25
Migrate Web App to v3 of the API


dadofsambonzuki:

dadofsambonzuki:
Comments done.

https://github.com/teambtcmap/btcmap.org/pull/409

dadofsambonzuki:
Boost WIP: https://github.com/teambtcmap/btcmap.org/pull/417

PR #421, opened 2025-10-08
refactor: merchant data to places v4 #181 #﻿422

escapedcat:
Partly fixes: #181
Fixes: #422 

> 
>   Goal: Complete v4 Places API migration for area pages (country/community views)
> 
>   Key Fixes:
> 
>   - Activity view - Fixed empty state (was using never-populated filteredElements instead of filteredPlaces)
>   - Stats view - Fixed race condition where enriched verification data wasn't loading before calculations
>   - Deleted places - Added filtering to prevent deleted locations from appearing
>   - MerchantCard - Fixed verification logic to use verifiedArr() for consistency across all components
>   - Merchant detail page - Removed store sync dependency for core cards, fixing flaky CI test (120s timeout)

### DIFF
--- a/src/routes/merchant/[id]/+page.svelte
+++ b/src/routes/merchant/[id]/+page.svelte
@@ -95,7 +95,6 @@
 		note = data.note;
 		hours = data.hours;
 		payment = data.payment;
-		// boosted and verified are now initialized reactively above
 		phone = data.phone;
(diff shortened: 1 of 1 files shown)


dadofsambonzuki:
Tested and looks good.

Shall we switch to `/places` now or later?

Issue #442, opened 2025-10-18
OG image preview not working on URLs BTC Map IDs


dadofsambonzuki:
https://btcmap.org/merchant/node:3793718458 gives is a nice OG image

https://btcmap.org/merchant/28628 does not.

dadofsambonzuki:
Clarification:

Both URLs should return the OG image.

escapedcat:
~I guess than we need to adjust the `og` API?~

Sorry, all good. Please have a look at the PR

PR #445, opened 2025-10-18
fix(merchant): correct OG image URL to use OSM Element ID #442

escapedcat:
Fixes: #442

It now doing this (again): https://api.btcmap.org/og/element/node:2489155938

### DIFF
--- a/src/routes/merchant/[id]/+page.svelte
+++ b/src/routes/merchant/[id]/+page.svelte
@@ -338,7 +338,7 @@
 		}
 	});
 
-	const ogImage = `https://api.btcmap.org/og/element/${data.id}`;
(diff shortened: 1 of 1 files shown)


escapedcat:
> Clarification:
> 
> Both URLs (OSM ID node/way/relation and BTC Map ID) should return the OG image for the place.

But then we need to adjust the API, no?

dadofsambonzuki:
Perfect! I see it working now. I was getting nothing returned sometimes. Perhaps the API times-out occasionally.

BTC Map ID: https://deploy-preview-445--btcmap.netlify.app/merchant/29285

OSM ID: https://opengraph.dev/panel?url=https%3A%2F%2Fdeploy-preview-445--btcmap.netlify.app%2Fmerchant%2Fnode%3A4585523015

Issue #444, opened 2025-10-18
Pin status refresh upon Boost


dadofsambonzuki:
Upon a successful boost, we should update the pin status (i.e. make it orange) and the boost info in the pin pop-up.

Currently this requires a page reload.

PR #446, opened 2025-10-18
fix(boost): update UI immediately after payment #444

escapedcat:
Fixes: #444

also sync latest data to localstorage when a merchant is being opened to i.e. sync latest comment count


dadofsambonzuki:
B E A Utiful 😍

PR #454, opened 2025-10-19
feat(map): add side drawer for merchant details

escapedcat:
~~Just a vibe-coded POC that could start the process of moving to something like googlemaps/airbnb as discussed in the bitcoin design call.~~

Finalized drawer approach. Currently only replacing the popup.  
Places list is a follow-up.

Notes from me:

- should the drawer be on the left side? -> did that
- what about the map buttons/lengend/infos? -> being covered for now (gmaps has those at bottom/right)
- future step would be to move a summary info of visible markers (need to define max) into the drawer -> follow-up task
- added a highlight to the clicked marker -> did that

#### Boosted
image
image

#### Not boosted
image
image

### ~Draft 2~

image
image
image

### ~Draft 1~

image

### DIFF
--- a/src/components/MerchantDetailsContent.svelte
+++ b/src/components/MerchantDetailsContent.svelte
@@ -81,7 +81,7 @@
 	<div class="border-t border-gray-300 pt-4 dark:border-white/95">
 		{#if merchant['osm:payment:onchain'] || merchant['osm:payment:lightning'] || merchant['osm:payment:lightning_contactless'] || merchant['osm:payment:bitcoin']}
 			<div class="mb-4">
-				<span class="block text-xs text-mapLabel dark:text-white/70">Payment Methods</span>
+				<span class="text-mapLabel block text-xs dark:text-white/70">Payment Methods</span>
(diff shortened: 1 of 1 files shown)


dadofsambonzuki:
I think the idea was to have a panel showing a number of place//merchant cards for the pins in view. Perhaps scrollable. This way you get a good view of all the places within the current map context without having to click on each pin,

Clicking on a specific place could highlight that card or bring it to the top or something?

escapedcat:
> I think the idea was to have a panel showing a number of place//merchant cards for the pins in view. Perhaps scrollable. This way you get a good view of all the places within the current map context without having to click on each pin,

Exactly, yes. That's what I tried to say with this:

> future step would be to move a summary info of visible markers (need to define max) into the drawer

But it's easier to start with this and then move on to create "the list"

> Clicking on a specific place could highlight that card or bring it to the top or something?

Not sure yet. 
Google does this:
image

Currently I think it's easier to have two view in the drawer:
1. detail
2. list

If you click on a marker or on an item in the list you get the detail view.

Does that make sense?

PR #496, opened 2025-11-16
feat(map): drawer latch for mobile

escapedcat:
Open points:

- [x] make it work nicely
- [x] DRY everything related to boosts, i.e. `handleBoost`

image

### DIFF
--- a/src/components/MerchantDrawerDesktop.svelte
+++ b/src/components/MerchantDrawerDesktop.svelte
@@ -1,128 +1,29 @@
 <script lang="ts">
-	import { places, boost, exchangeRate, resetBoost } from '$lib/store';
+	import { boost, exchangeRate, resetBoost } from '$lib/store';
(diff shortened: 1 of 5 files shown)


bubelov:
Unsure what's going on here but this error appears every time I try to refresh a page, it also stops displaying icons. "deep refresh" works though, so probably something related to cache?

Screenshot From 2025-11-18 15-18-25

escapedcat:
> Unsure what's going on here but this error appears every time I try to refresh a page, it also stops displaying icons. 

~~Can you try in a private tab?~~

> "deep refresh" works though, so probably something related to cache?

Hard to say. I had some of these but not only on this branch. Was wondering if this is related to slow connection sometimes.

bubelov:
I checked that other PR and the issue persists. Here is a request log and a console output:

Screenshot From 2025-11-19 10-25-14


Screenshot From 2025-11-19 10-26-03

PR #502, opened 2025-11-18
refactor: switch dashboard to REST endpoint

bubelov:
Main changes:

- Elements renamed to merchants, atms/exchanges are excluded
- Days since verified chart is gone as it's hard to understand and it's more of an internal metric
- The API is already returning echange data and charts, I'll add them in future commits to this PR
- Some numbers are slightly out of sync for historical reasons, new reports will use new definitions of exchanges and merchants so this issue will resolve itself with time as we generate more new and more precise daily reports

### DIFF
--- a/src/routes/dashboard/+page.server.ts
+++ b/src/routes/dashboard/+page.server.ts
@@ -1,40 +1,26 @@
 import type { PageServerLoad } from './$types';
+import { error } from '@sveltejs/kit';
 
-export const load: PageServerLoad = async () => {
-	try {
-		const response = await fetch('https://api.btcmap.org/rpc', {
-			method: 'POST',
-			headers: {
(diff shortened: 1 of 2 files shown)


escapedcat:
Code lgtm

This difference is intentional? (i.e. number-count and verified locations)

|  PR | PROD |
|:-:|------------------------|
| image  | image  |

bubelov:
Yes, the old numbers included ATMs (amenity = `atm`) and exchanges (amenity = `bureau_de_change`). Merchants is basically everything except for that, so you end up with a slightly lower number.

We will also show exchange counts in the same way, although I'm not sure if we should have exchange charts since it's not the main focus of BTC Map. There are much better places for exhange maxis, such as https://coinatmradar.com/ and friends.

PR #503, opened 2025-11-18
fix(sync): retain CDN data when API update fails on first load

escapedcat:
Previously, when a first-time user loaded the app and the CDN download succeeded but the API update check failed, all downloaded data was discarded and the map would not load. This resulted in a poor user experience despite having valid baseline data.

Now, when the API fails to fetch recent updates, the app continues with the existing data (either from cache or freshly downloaded from CDN) and shows the same user-friendly toast message in both scenarios.

This ensures the map always works if baseline data can be downloaded, with graceful degradation when real-time updates are unavailable.

### DIFF
--- a/src/lib/sync/places.ts
+++ b/src/lib/sync/places.ts
@@ -146,6 +146,17 @@ export const elementsSync = async () => {
 							}
 						});
 
+						// Validate parsed data
+						if (!Array.isArray(placesData) || placesData.length === 0) {
+							console.error('CDN data parsing failed:', {
+								isArray: Array.isArray(placesData),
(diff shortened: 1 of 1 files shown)


bubelov:
The code looks good

I noticed that our CDN snapshot was terribly out of date, I guess it's time to set up a reliable systemd scheduler: https://github.com/teambtcmap/btcmap-api/issues/69

Are you using the same list of fields listed here or is there anything extra the website expects from a snapshot? https://github.com/teambtcmap/btcmap-api/blob/master/docs/rest/v4/sync.md#cdn-powered-static-snapshots

Issue #495, opened 2025-11-15
Buttons on communities page need padding


dadofsambonzuki:
![Screenshot_20251115-183748.png](https://github.com/user-attachments/assets/58d6fe59-82d2-4631-bedc-20689a14e5ea)

escapedcat:
I guess on mobile they should just be 100% width like the "leaderboard" one

PR #507, opened 2025-11-20
fix: Adjust padding on community buttons mobile view

jk1551:

Fixes: https://github.com/teambtcmap/btcmap.org/issues/495

The change adjusts the width to be handled with w-full instead of mx-auto due to margin not being set smaller than md

**Screenshots**

Before:

IMG_0020

After:

IMG_0019

**Additional context**


escapedcat:
💅

Issue #490, opened 2025-11-11
Site freezes for around 5 seconds on first load


SubatomicPlanets:
**Describe the bug**
When I go to the site btcmap.org, it works properly for around a second, but then it "freezes". I can still scroll, but I cannot click any buttons. Hovering over menus or buttons has no effect, like a color change. After around 5 seconds, everything is normal again. This only happens on the first load.

**To Reproduce**
Delete cookies and cache or just use a private/incognito tab and go to btcmap.org

**Additional context**
I used Firefox on my Windows PC, but I also tested it on Safari on an iPhone 12. Both freeze on the first load.

escapedcat:
Uhm, weird.

Can you try a private tab?
What exact browser (version) and device do you use?
Can you try chrome/brave?

SubatomicPlanets:
@escapedcat I use the newest Firefox version on Windows (version 145.0.1). It also still happens in private tabs. On my phone (using samsung internet which is based on chromium) it also freezes.

escapedcat:
Would it be possible to share a recording of that?

escapedcat:
Did another change, please try again:
https://deploy-preview-508--btcmap.netlify.app/

escapedcat:
Try again please ;) Eventually we might figure this out

SubatomicPlanets:
@escapedcat Yes, it works! Thanks for fixing it!

escapedcat:
Thanks for your debugging help!

PR #508, opened 2025-11-20
fix(sync): await IndexedDB operations to prevent UI freeze #490

escapedcat:
Maybe fixes: #490


bubelov:
So, my method is to open a link in a private tab and to toggle the "Stats" menu. It still freezes for 10 seconds every time

FF profiler blames `find` and `filter` calls, as far as I understand:

Screenshot From 2025-11-20 12-30-14

PR #518, opened 2025-11-25
feat(map): add merchant list panel for desktop at high zoom levels

escapedcat:
# DESKTOP ONLY

Offer a sidebar panel with nearby merchants. When zoomed to a certain level the list will be fille. Features on-demand data fetching with skeleton loading states and pan-to-merchant when selecting from list.

### draft 5

- adjusted nearby to extend the area based on zoom
- from 50 places and below the list becomes usable

#### too far

image
image

#### close enough

image
image

#### close enough with grouped markers

image

#### expanded

image

image

#### expanded with detail
image

image

#### collapsed

image
image

### draft 2

image

image

### draft 1

**Screenshots**

image
image
image


bubelov:
Is that gap on the left side intended, in case a nearby panel is hidden?

Screenshot From 2025-12-01 13-13-06

Issue #469, opened 2025-10-24
Boost for one month always 5000 sats


beinardus:
While the conversion on screen is correct; showing the correct conversion from sats to fiat $, when the invoice is scanned, it is exactly 5000 sats.

<img width="375" height="158" alt="Image" src="https://github.com/user-attachments/assets/1727fe69-07b9-46ca-adb2-bc3e62245941" />

![Image](https://github.com/user-attachments/assets/b46d0c75-4726-434d-bfec-61cc72e4a50f)

beinardus:
Verified with `lncli decodepayreq`.
"num_satoshis": "5000" in the JSON of the invoice vs 4502 below the QR code.

bubelov:
Our pricing is in sats, not in USD (API knows nothing about fiat and doesn't have exchange rate data), so clients have 2 options:

1. Just show sats and let user wallets handle the conversion in user-preferred/preset currency (ideal imo)
2. Try to force Dollars on everyone or try to figure out user preferred currency based on browser locale, ip address, etc, which is better for localization but kind of retarded and pretty complex to implement and maintain. Ideally we would need a settings screen with a currency picker, etc

I'm using the first approach on Android and I think it would work just fine for a web app too

dadofsambonzuki:
Happy to switch to sats native/first.

PR #523, opened 2025-11-28
feat: boost amount in sats instead of fiat #469 

trigger-xyz:
Fixes: #469 

Boost amounts are changed from fiat to sats

I have not removed the sats price, but displayed it for information. It might be usefull for some people to know the corresponding value in dollars...
I'm not 100% convinced, if you still think fiat should disapear from the website, I'll update my PR!

I would like to try another UI for the "prefered option" too. I did let the star icon for the moment, but can suggest something else in this / a future PR.

**Screenshots**

Old version : 
image

New version : 
image

### DIFF
--- a/src/components/Boost.svelte
+++ b/src/components/Boost.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
 	import { invalidateAll } from '$app/navigation';
 	import CloseButton from '$components/CloseButton.svelte';
-	import { boost, exchangeRate, resetBoost, lastUpdatedPlaceId } from '$lib/store';
+	import { boost, resetBoost, lastUpdatedPlaceId } from '$lib/store';
 	import OutClick from 'svelte-outclick';
 	import { fly } from 'svelte/transition';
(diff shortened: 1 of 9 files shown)


trigger67:
I removed the exchangeRate from everywhere, and reworked a bit the UI.

Old version : 
image


New version : 
image

I changed the elements order, having the product (duration) on top with a small text, and a bigger-bolder price under it feels better.

For the preferred version, I didn't change anything, as I was not sure about adding a "recommanded" text. Also I also thought about highlighting it, but I was not sure about the color, and the fact that we already have a different color on hover/selection could make that feel strange.

escapedcat:
> Would be super interesting to A/B test this to see if there is any dollar-sats bias when choosing to spend.

We do not have a A?B testing setup currently, right? OR did we ever do such tests before?
If nor I suggest to merge this and wait for feedback :P 

Could look into A/B testing though.

trigger67:
Linting issue is fixed!

PR #524, opened 2025-11-28
fix: add cursor-pointer in all button

trigger-xyz:
In many places, there is no cursor-pointer when hovering buttons.

### DIFF
--- a/src/app.css
+++ b/src/app.css
@@ -15,6 +15,14 @@
 	::selection {
 		@apply bg-[#53c5d5] text-white;
 	}
+
+	button {
+		cursor: pointer;
+	}
+
+	button:disabled {
+		cursor: not-allowed;
+	}
 }
 
 @layer utilities {


escapedcat:
An issue with this might be that disabled buttons now also have a pointer, which might be confusing?

image

Compare these
- https://deploy-preview-524--btcmap.netlify.app/merchant/1158
- https://btcmap.org/merchant/1158


I did some quick research and asked Claude regarding the links I found:

- https://ux.stackexchange.com/questions/105024/why-dont-button-html-elements-have-a-css-cursor-pointer-by-default
- https://a11y-solutions.stevenwoodson.com/solutions/semantics/buttons-and-links/
- https://stackoverflow.com/questions/4121854/is-it-wrong-to-use-the-hand-cursor-for-clickable-items-such-as-buttons


>   Key Findings
> 
>   The UX Debate
> 
>   | Against cursor: pointer                  | For cursor: pointer                          |
>   |------------------------------------------|----------------------------------------------|
>   | W3C: pointer = links only                | Modern flat design weakens button affordance |
>   | Buttons already look clickable           | Users don't care about semantics             |
>   | Desktop apps use default cursor          | Even Apple/MS use pointer on their sites     |
>   | A11y: helps distinguish links vs actions | "Don't make me think" principle              |
> 
>   The Disabled Button Problem
> 
>   You're absolutely right - the PR's current code:
> 
>   button {
>       cursor: pointer;
>   }
> 
>   Would apply pointer cursor to disabled buttons too, which is bad UX. Disabled buttons should use cursor: not-allowed or default cursor.
> 
>   Should be:
> 
>   button:not(:disabled) {
>       cursor: pointer;
>   }
> 
>   Or even more complete:
> 
>   button {
>       cursor: pointer;
>   }
> 
>   button:disabled {
>       cursor: not-allowed;
>   }
> 
>   Updated Review
> 
>   ⚠️ Request Changes:
> 
>   1. Technical bug: Doesn't handle disabled state
>   2. Debatable UX: The change itself is contested in UX community - not clearly "right"
> 
>   The PR author should either:
>   - Add :not(:disabled) selector, or
>   - Add explicit button:disabled { cursor: not-allowed; } rule

trigger67:
Hey, thanks for the links, I neved realized that this could be a complex topic ux wise!

After reading through everything, my opinion would be to put a pointer on **everything that is clickable.**
It seems to be the case in every websites I visit, and it's what I've always done. I like to challenge the status quo, but I think it's a standard and users are really used to that.
Personnaly I was disturbed not having pointers on the menu and the buttons.

For the disabled buttons you are right! I rushed the think a little bit, thinking that in any case css will be refactored once migrated to tailwind v4, but let's do it properly :)

I'll update the PR in a minute.

escapedcat:
Thanks.
I do not have very strong opinions on this, just need to make sure that it's not adding new confusion :P

>  css will be refactored once migrated to tailwind v4

We're already [on v4](https://github.com/teambtcmap/btcmap.org/blob/main/package.json#L82C4-L82C28) btw

Issue #526, opened 2025-11-30
Refactor components


trigger67:
There are more than 80 components in /src/components.

I would recommand defining a logic and organizing them. It's way easier for developers to find stuff they are looking for, to avoid duplications, and to build a proper component library / design system.

There are many ways to do that. First I would move the "page specific" components, the ones that are used only once, next to their corrsponding page. Example:

└── routes/
    ├── about-us/
    │   ├── _components/ 
    │   │   └── AboutCommunity.svelte
    │   │   └── AboutMerchant.svelte
    │   ├── +page.svelte

Then for the remaining components, I would create a specific folder for bases components and for composites.
- A bases component in a very simple element: button, input, loader, select, etc.
- Composites are more complex.

Then maybe one folder for each "big feature" to put some "feature related" components.
And if needed a "pattern" folder for big reusable components.

I enjoy refactoring, let me know if you want me to open a PR, maybe start with a first one for "page specific" components :)

 ** sorry wrong tag

PR #542, opened 2025-12-03
Refactor components

trigger-xyz:
#526 

Let me know if that's good, if yes I'll do all other "page specific" components in the next PR!


trigger67:
Hmmm strange stuff happened with prettier there, but yeah fixed thx.

PR #557, opened 2025-12-08
feat(map): enable swipe-to-collapse from expanded drawer content #235

escapedcat:
Fixes: #549

Allow users to swipe down from anywhere in the expanded merchant drawer (not just the handle) to collapse to peek view. Uses scroll-aware gesture detection: only initiates collapse when content is scrolled to top and user drags downward, preserving normal scroll behavior otherwise.


dadofsambonzuki:
Much better 😍

PR #583, opened 2025-12-16
Further tagger leaderboard enhancements

dadofsambonzuki:


src/routes/leaderboard/+page.svelte
```diff
@@ -295,7 +295,7 @@
 		</div>
 	</div>
 
-	<div class="mx-auto w-10/12 space-y-10 xl:w-[1200px]">
+	<div class="mx-auto w-[95%] space-y-10 xl:w-[1200px]">
```
escapedcat: Is there no way we can use an existing grid value from tailwind? Using individual values seems like the wrong way to go

Issue #576, opened 2025-12-14
Add Tagger Onboarding Page/Form


dadofsambonzuki:
We need an easy way for new taggers to be onboarded.

We now have a [new Gitea Issue type](https://gitea.btcmap.org/teambtcmap/btcmap-data/issues?labels=1394) where the onboarding can be managed - access to systems, orientation, mentoring/shadowing.

We now need a form to create these Issue types in a similar way to add/verify locations.

We should also use this as an opportunity to switch add/verify issue creation to use IDs and not name as they currently do so that we can transition adds and verifies to the `type/` issue namespace.

escapedcat:
Can you elaborate "the `type/` issue namespace"?

dadofsambonzuki:
> Can you elaborate "the `type/` issue namespace"?

Just a way to force specific label selections in Gitea. e.g. you can set it to only allow one label per group. So either `type/add-location` or `type/verify-location` or `type/tagger-onboarding` etc.

escapedcat:
What info do we need? Only a name? No too familiar with the backoffice yet

dadofsambonzuki:
Initial test looks good.

I'll try re-naming labels later and see if anything breaks!

PR #584, opened 2025-12-16
feat(tagger): add tagger onboarding form #576

escapedcat:
Fixes: #576 

Add new page at /tagger-onboarding for prospective taggers to submit their name and email. Creates a Gitea issue with type/tagger-onboarding label. Added navigation link in Maintain dropdown.

### DIFF
--- a/src/components/area/AreaTickets.svelte
+++ b/src/components/area/AreaTickets.svelte
@@ -1,23 +1,25 @@
 <script lang="ts">
 	import InfoTooltip from '$components/InfoTooltip.svelte';
 	import OpenTicket from '$components/OpenTicket.svelte';
+	import { GITEA_LABELS } from '$lib/constants';
 	import type { Tickets } from '$lib/types.js';
 	import { errToast } from '$lib/utils';
 
+	import type { GiteaLabel, GiteaIssue } from '$lib/types';
+
 	export let title: string;
 	export let tickets: Tickets;
-	import type { GiteaLabel, GiteaIssue } from '$lib/types';
 
 	$: filteredTickets = tickets === 'error' ? [] : tickets;
 
 	$: add = filteredTickets.filter((issue: GiteaIssue) =>
-		issue.labels.some((label: GiteaLabel) => label.name === 'location-submission')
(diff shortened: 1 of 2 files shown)


src/lib/gitea.ts
```diff
@@ -96,92 +95,26 @@ export async function getIssues(
...
-		return [];
-	}
-}
+	const response = await axios.post(
+		`${env.GITEA_API_URL}/api/v1/repos/teambtcmap/${repo}/issues`,
+		{ title, body, labels: labelIds },
+		{ headers }
+	);
 
-async function createLabel(name: string): Promise<number | null> {
```
dadofsambonzuki: Don't we need this functionality to create the area labels for add-location and verify-location?

We will still need to query those (and create where necessary) by name. I think just the types we can do by ID.

dadofsambonzuki:
Tested and looks good. Bot added to the right teams in GT so issues being raised properly in the infra repo.

If you're happy with the changes then merge.

dadofsambonzuki (commented):
We're back to label names now (albeit with the /type namespace) - full circle!

https://github.com/teambtcmap/btcmap.org/pull/584/commits/27366882364473b18b4ec013484ef8d8f056511a

Either way is fine I guess, each has benefits/tradeoffs.

We'd have to make those label name changes when we go live, else we'll have a temp label hardfork (if it creates on the flay like with area labels) or errors with no label found. IDs prevent this.

API App uses IDs.

I saw the CoPilot recomendation - but I think it's fine to hardcode important labels like type as we're more likely to change the namespace and name, than ID.

src/lib/gitea.ts
dadofsambonzuki: Ah, ok. All good.

escapedcat:
> We're back to label names now (albeit with the /type namespace) - full circle!
> 
> [2736688](https://github.com/teambtcmap/btcmap.org/commit/27366882364473b18b4ec013484ef8d8f056511a)

Good point, adjusted!

escapedcat:
🤞

PR #610, opened 2025-12-24
refactor(map): control leaflet default styling

escapedcat:
Use defaults instead of individual styling to reduce code. Also this styling seems nicer.
Just an idea.

### PROD
image
image

### PR

image

image


dadofsambonzuki:
Christmas Day merges like a boss. 🤜🤛

escapedcat:
A click of a button is still within the boundaries :P

PR #742, opened 2026-02-25
feat(i18n): translate dashboard page title

Nackoo2000:


.github/actions/setup/action.yml
```diff
@@ -15,7 +15,7 @@ runs:
   steps:
     - uses: actions/setup-node@v4
       with:
-        node-version: lts/*
+        node-version: '20'
```
escapedcat: Why?

PR #785, opened 2026-02-27
feat(map): add language parameter support for embedded website (all pages)

Nackoo2000:
- Remove redirect from /map?language=X to communities map so main map can use the param
- Read language from URL and set locale when en, pt-BR, or bg (persist to localStorage)
- Preserve search params in updateMapHash so language stays in URL when zooming/panning

Enables embedded maps (e.g. btcfamily.eu) to display in user's preferred language: https://btcmap.org/map?lat=42.7&lon=25.5&zoom=6&language=bg

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve URL query parameters (e.g., language) when updating the hash so search params are not lost.
  * Removed server-side redirect behavior that previously acted on the language query parameter.

* **New Features**
  * Client-side language initialization from URL with persistence to local storage.
  * Added explicit validation of supported locales.

* **Chores**
  * Renamed communities language query parameter to "communityLang".

### DIFF
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -30,8 +30,8 @@ import { isSupportedLocale } from "$lib/i18n";
 import { page } from "$app/stores";
 
 // Apply language from URL param site-wide (e.g. /map?language=bg for embedded maps).
-// Exclude /communities/map where language filters communities by area.tags.language.
-$: if (browser && $page.url.pathname !== "/communities/map") {
+// On /communities/map, ?communityLang= filters communities; ?language= sets UI locale.
+$: if (browser) {
 	const langParam = $page.url.searchParams.get("language");
 	if (langParam && isSupportedLocale(langParam)) {
 		locale.set(langParam);
--- a/src/routes/communities/map/+page.svelte
+++ b/src/routes/communities/map/+page.svelte
@@ -46,7 +46,7 @@ let communitiesLoaded = false;
 const communityQuery = $page.url.searchParams.get("community");
 
 // allow to view map with only certain language communities
-const language = $page.url.searchParams.get("language");
+const communityLang = $page.url.searchParams.get("communityLang");
 
 // allow to view map with only certain org communities
 const organization = $page.url.searchParams.get("organization");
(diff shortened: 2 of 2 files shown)


Nackoo2000:
Thank you, man, for the help @escapedcat. Now I am ready to use the map in my website - I will check from time to time to update the translations when you add new features. This process increased my understanding and skills on how open source works (still a noob, but it was eye-opening)

escapedcat:
@Nackoo2000 thanks for doing this and being patient enough! 👏

Nackoo2000:
> > > Looking at the PR, I'd flag one potential concern:
> > > In +layout.svelte, the language param is read on every page, not just /map. So visiting /?language=bg would also apply Bulgarian to the entire site, not just embedded maps.
> > > If that's intentional (allow language override anywhere for embeds), then it's fine. But if you want it scoped only to map embeds, it should be in src/routes/map/+layout.svelte instead of the root layout.
> 
> > PR #785 would add the locale-setting logic. If it goes to root layout (+layout.svelte), it would apply to all pages when someone visits with ?language=bg.
> > Better approach: Put it in src/routes/map/+layout.svelte (or src/routes/communities/map/+layout.svelte) so it's scoped only to map pages, not the whole site.
> > That's the fix needed if you want to move forward with this feature.

I want the website to be translated everywhere not just maps - I may add the form in a separate page for an easier UX of the users on my website or anything else that might be better I do not want only the maps to have this option.

dadofsambonzuki:
Makes sense.

escapedcat:
> But you're right — renaming the community filter param would be cleaner:
> | Current | Cleaner alternative |
> |---------|---------------------|
> | /communities/map?language=de (filter) | /communities/map?communityLang=de |
> | ?language=bg (locale) | ?language=bg |
> This way:
> - ?language= always means UI locale (works everywhere)
> - ?communityLang= filters communities
> Downside: Breaking change for anyone currently using ?language= to filter communities.

We're aware of the downside, but it makes it cleaner on the long run. If people currently use this they can adjust the link.

escapedcat:
> And should be proper en-GB, en-US, bg-BG?

Could be adjusted in a follow up as well I guess

escapedcat (changes requested):
See comments

escapedcat (approved):

PR #813, opened 2026-03-13
perf(leaderboard): speed up to bring it to sub-second expected latency

bubelov:
This change switches the leaderboard to a query we're using for our monthly reports. It is much faster than using cache or assembling data piece by piece.

It also fixes the issue when the default range is 3 months but default data actually covers 12 months.

## Summary by CodeRabbit

* **Localization**
  * Updated Russian UI text across navigation, dashboard, and leaderboard for clearer terminology and grammar.

* **Leaderboard**
  * Default reporting period changed from 12 to 3 months.
  * Switched to a simplified REST data fetch for leaderboard data.
  * All editors are now included in leaderboard rankings (no more exclusions).

### DIFF
--- a/src/routes/leaderboard/+page.svelte
+++ b/src/routes/leaderboard/+page.svelte
@@ -141,7 +141,7 @@ const normalizeUsers = (users: TopEditorItem[]): TaggerRow[] => {
 };
 
 $: {
-	if (data?.rpcResult?.users?.length) {
+	if (Array.isArray(data?.rpcResult?.users)) {
 		const normalizedUsers = normalizeUsers(data.rpcResult.users);
 		leaderboardRows = normalizedUsers;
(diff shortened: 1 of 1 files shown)


escapedcat:
@bubelov can you have a look at the two qodo comments?

> 2. Tip url treated lightning 

Not sure about this one

> 1. Missing excludeleader export 

If `excludeLeader` is still being imported in other views it should be cleaned up

bubelov:
Tip links look fine, the assumption that server would return something http/s prefixed is weird, it displays tip_url filed as is:

https://api.btcmap.org/v4/top-editors?period_start=2026-02-13&period_end=2026-03-13

tip_url	"lightning:rockedf@coinos.io"

Which is indeed a valid URL and I see it in browser:

Screenshot From 2026-03-13 18-07-31

PR #938, opened 2026-04-22
feat: activity card save

escapedcat:
Not sure about the design

## User feed
image

## Area feed
image

### DIFF
--- a/src/components/activity/ActivityCard.svelte
+++ b/src/components/activity/ActivityCard.svelte
@@ -2,6 +2,7 @@
 import Time from "svelte-time";
 
 import Icon from "$components/Icon.svelte";
+import SaveButton from "$components/SaveButton.svelte";
 import { type ActivityItem, dotColor } from "$lib/activity";
 import { _ } from "$lib/i18n";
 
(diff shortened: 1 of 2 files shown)


dadofsambonzuki:
This would add individual activity items to the 'My Saved' page?

escapedcat:
This will allow the user to save individual places from their stream.
I.e. following Spain, seeing a new cafe opened, and save that new cafe to their places.

Does this make sense?

dadofsambonzuki:
Hmm, maybe 'Save Place' then.

Lot's of repetition and UX noise for an edge use case though as per your OP.

Fine for now, can optimise later. Maybe a hover over, but then that would likely hide it from most people.

The alternative flow is they click on a Place and Save it from there as usual. Perhaps that's intuitive enough?

escapedcat:
> Hmm, maybe 'Save Place' then.

Can do

> Lot's of repetition and UX noise for an edge use case though as per your OP.

Just an idea to drive engagement. We can also not add it. All good.

> Fine for now, can optimise later. Maybe a hover over, but then that would likely hide it from most people.

Hover wouldn't work on mobile. But yeah, design can be optimized, this wa sjust the most simple thing to do.

> The alternative flow is they click on a Place and Save it from there as usual. Perhaps that's intuitive enough?

Yeah, not sure. Might be nice if users can just directly add new places instead of opening each place one by one they see in their feed?

dadofsambonzuki (approved):

PR #940, opened 2026-04-22
feat(map): drawer save button

escapedcat:
image
image


dadofsambonzuki:
Why is the navigate button in the selected style by default?

Same on prod btw.

escapedcat:
Will look into it, has always been like this I guess. Apart from that is this PR ok?

dadofsambonzuki (approved):

escapedcat:
Claude:

>   So it's an intentional "primary CTA" pattern — Navigate is always filled because directions is the action users most commonly want on a place card (matches Google Maps). Edit/Share/Comments are secondary, outline. Nothing to fix.
> 
>   Our new Save button's filled-when-saved / outline-when-unsaved is a toggle-state indicator — a different semantic than Navigate's always-primary role, so the visual overlap (both can be filled primary) happens to coexist cleanly:
>   Navigate = "go here", Save = "saved here."

PR #947, opened 2026-04-25
feat(addLocation): manual form coords #50

### LINKED ISSUE #50: Allow entry of coordinates on the noob form.
Allow users to enter raw coordinates on the noob form.

The marker should show the location on the map once both fields have been filled.

escapedcat:
Fixes: #50

image
image


dadofsambonzuki:
I think a better UX would just be to make the existing co-ords fields edible when in advanced mode.

Also enables switching back to normal mode as currently once you've entered coords, you can't then move the pin.

dadofsambonzuki (approved):

PR #970, opened 2026-05-04
fix(i18n): nl merchant button layout

escapedcat:


LighthouseFR (approved):
I see a few lines that don't appear in my nl.json file, so I have no idea how they were modified. I do see that apparently certain translations were too long and had to be shortened? 
I'm glad you managed to fix it.

PR #991, opened 2026-05-10
feat(addLocation): nominatim address search

escapedcat:
⏺ ## Summary

  Adds an address search to the **Add Location** form. Contributors can type a mailing address instead of panning the map to   pick lat/long — addresses feedback that coordinates are unfriendly for non-technical users.

  ## How it works

  - Type an address → click **Search** (or press Enter) → results from [Nominatim](https://nominatim.openstreetmap.org/)  appear
  - One result → auto-selected, dropdown skipped
  - Multiple results → small dropdown; pick one
  - Selecting a result flies the map to the spot, drops the marker, and auto-fills the optional Address field
  - Subsequent searches refresh the Address field — unless the user has typed there themselves, then it stays locked to  their input
  - Existing flows (map-click, advanced lat/long entry) are unchanged

  ## Rate-limit handling

  Requests go straight from the browser to `nominatim.openstreetmap.org`. Identification relies on the `Referer: btcmap.org`   header (browsers can't set a custom `User-Agent` cross-origin). The loading state prevents back-to-back clicks; network  round-trip naturally rate-limits. An 8s axios timeout caps the worst case if Nominatim hangs.

  If abuse becomes a real problem, the right answer is a server-side proxy (`/api/geocode`) with IP-level rate limiting and  a proper `User-Agent`. Intentionally out of scope here — ship the UX win first, add the backend layer once we see real  traffic.

### Input location

image

### See possible results

image

### After selecting result
image

## Summary by CodeRabbit

* **New Features**
  * Address search/autocomplete added for selecting locations, including keyboard and result handling; selections update the map and form fields.

* **Refactor**
  * Hidden anti-bot field placeholders standardized to a fixed English string across forms.
  * i18n locale files updated to replace honeypot keys with captcha-related strings and formatting improvements.

* **Tests**
  * Added tests covering the geocoding/address-search backend behavior.

* **Documentation**
  * i18n guideline added clarifying honeypot placeholder handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teambtcmap/btcmap.org/pull/991)


dadofsambonzuki:
Nice!

Some comments:

1. I think we should remove the countdown and subtext, keep the manual search. If people spam it we can do something about it. Maybe add 'Search for address' OR 'Select location on the map' ?

2. The 'Address (Optional) ' field does not update beyond the first address searched for and selected.

3. Prbly from the last change, but we should also remove the reference to Google Maps from the advanced section.

dadofsambonzuki (approved):

Issue #972, opened 2026-05-05
Nav and home icon jitter


dadofsambonzuki:
**Describe the bug**

When navigating away from and to the home page, the whole nav section and home icon jitters slightly

**To Reproduce**

Click a menu item, then click the home icon.

**Expected behavior**

Nav and icon remain stationary.

**Screenshots**

**Additional context**

escapedcat:
Can't reproduce this on desktop (also not with devtools mobile) or on android. Got a video?

dadofsambonzuki:
I've narrowed it down. It only happening on pages where there is no vertical scrolling. The scroll bar only disappears on short pages (home, tagger-onboarding ) viewed on big screens, hence the difficulty to reproduce.

escapedcat:
Ah, yeah, kinda typical thing really. Wouldn't care to fix that.

escapedcat:
There might be a way to handle this a bit nicer though

PR #994, opened 2026-05-12
fix(layout): reserve scrollbar gutter to prevent header jump #972

escapedcat:
Adds scrollbar-gutter: stable on html so the layout stays put when navigating between pages with and without a vertical scrollbar. Overrides back to auto for /map and /communities/map via svelte:head so those fullscreen leaflet views still fill the viewport edge-to-edge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented layout shifting caused by scrollbar appearance and disappearance across the application
  * Improved fullscreen map view stability with consistent viewport spacing

[Review Change Stack](https://app.coderabbit.ai/change-stack/teambtcmap/btcmap.org/pull/994)

### DIFF
--- a/src/app.css
+++ b/src/app.css
@@ -11,6 +11,7 @@
 	html {
 		font-family: Manrope, sans-serif;
 		@apply text-base;
+		scrollbar-gutter: stable;
 	}
 
 	::selection {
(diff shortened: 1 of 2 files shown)


dadofsambonzuki:
Tested an lgtm!

Issue #996, opened 2026-05-12
Map rotation 


dadofsambonzuki:
Enable two-finger gesture rotation on map with compass.

escapedcat:
<img width="1182" height="733" alt="Image" src="https://github.com/user-attachments/assets/3d4baf5e-c6a6-404a-8504-60b269b634b0" />

<img width="1149" height="820" alt="Image" src="https://github.com/user-attachments/assets/ea969d1e-2d98-4d01-93b6-ac22889e761e" />

Computer says no

Edit:
See dev env on mobile: https://github.com/teambtcmap/btcmap.org/pull/1002

Question:
You have examples of OSS/leaflet projects which have this working?

dadofsambonzuki:
What is this magic?!

escapedcat:
>> What is this magic?!

> give me a html presentation of the how we got here and what the reults of the spike are

(see edits above as well)

dadofsambonzuki:
Yup, leaflet and maplibre clash here.

For us, Leaflet came first, then I think Maplibre bindings were added for vector support as it was the lowest effort. Leaflet doesn't support vector tiles natively.

The other option is to wholesale swap Leaflet for Maplibre natively as both support rotation, just not together.

Worth weighing the two frameworks and picking one imo.

https://maplibre.org/maplibre-gl-js/docs/guides/leaflet-migration-guide/

escapedcat:
Looking into it 🤯

PR #1003, opened 2026-05-13
feat(map): switch to maplibre native

escapedcat:
Re-writing all ze maps to solve #996 🤞

  - Map switched from Leaflet to MapLibre — smoother pan/zoom, rotation + pitch, modern WebGL rendering
  - /merchant/[id], /add-location, /communities/map, area + saved-place maps all migrated too — consistent look across the site
  - Browsers without WebGL get an explanatory fallback instead of a blank map
  - Three selectable basemaps via the layer picker (OSM, Carto Light, Carto Dark)
  - Dark-mode treatment for the control pill, layer icon, and community popup
  - Boosted listings stay individually visible at every zoom (no longer absorbed into clusters) and correctly show their comment-count badge, saved-bookmark badge, and place-name label
  - Community / country / saved-places maps now cluster pins at low zoom so dense cities aren't a wall of overlapping pins
  - Map pins filter with the panel — picking a category or running a search narrows the map's pins to match the list
  - Boosting a place or leaving a comment updates that pin on the map immediately
  - Searching a place zooms close enough to actually see the pin instead of stopping at a cluster
  - "Locate me" eases straight to your position instead of flying a parabolic arc
  - Returning visitors land back where they left off; first-time visitors get an IP-derived starting view
  - Theme toggle preserves your pan/zoom
  - Sync errors surface as a toast instead of silently leaving the map empty
  - Bad URL coordinates show a toast instead of silently defaulting
  - Embed contract preserved: `?lat=&long=` (single point + bounds-pair) still works; deep-linked street-level zoom (21) works again everywhere
  - Scale bar restored; "Support BTC Map" link restored in the attribution
  - Navigating community → community shows the new area's pins immediately; the merchant drawer closes if its place isn't in the new area
  - Multipart communities render the full boundary (previously only the first piece showed)
  - /communities/map's basemap picker is properly styled now (was unstyled fallback)
  - Search-radius math handles the antimeridian — searches around Fiji / NZ / eastern Russia no longer ask the API for half the planet
  - Map doesn't repeat horizontally at low zoom

## Summary by CodeRabbit

* **New Features**
  * Full migration from Leaflet to MapLibre GL with selectable basemaps, new map controls (basemap picker, boost toggle, nav buttons, data refresh), improved pin/sprite rendering, URL hash viewport sync, and refined hover/preview behavior.

* **Refactor**
  * Map-related UI and interactions reimplemented for MapLibre, simplifying lifecycle and style switching.

* **Bug Fixes**
  * More consistent cluster/marker rendering, theme switching, and popup/cleanup handling.

* **Tests**
  * End-to-end tests updated for MapLibre DOM and WebGL rendering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teambtcmap/btcmap.org/pull/1003?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


dadofsambonzuki:
Looks and feels great. Rotate and pan on mobile work well. I also like the more granular zoom levels all round.

And the merchant deep links on the map are super cool too!

Feedback:

- Rotation gestures seem weird on desktop. Shift-click-mousemove has the map rotating the opposite way I would expect. Clicking-hold-mousemove on the compass icon is a ride!
- Feels like place name labels should appear at a higher zoom. Maybe 10? I guess clustering would overide this.
- Basemap switcher has dark background in dark mode.

escapedcat:
> Looks and feels great. Rotate and pan on mobile work well. I also like the more granular zoom levels all round.

🙌 


> And the merchant deep links on the map are super cool too!

Which ones? I assume everything is like before? :D

 
>     * Rotation gestures seem weird on desktop. Shift-click-mousemove has the map rotating the opposite way I would expect. Clicking-hold-mousemove on the compass icon is a ride!

Will check, I assume currently everything as it comes default OOTB


>     * Feels like place name labels should appear at a higher zoom. Maybe 10? I guess clustering would overide this.

Should be same as it is on prod currently. But can be adjusted.


>     * Basemap switcher has dark background in dark mode.

You mean that select or that light/dark-theme switcher?



Sound like this is mainly fine. Will add missing changes step by step via follow up PRs into this one.

dadofsambonzuki:
Maybe rotations on desktop is user error. If you use mouse left/right, rather than mouse up/down it's fine.

dadofsambonzuki:
> You mean that select or that light/dark-theme switcher?

No, the tile selector.

image

dadofsambonzuki:
Sooooo much better!

Only nit is the close button on the community map looks a bit off.

Screenshot From 2026-05-27 11-04-15

dadofsambonzuki:
A more general thoughts was around any impact to apps that have integrated via WebView/iFrame.

Here's what GPT5.5 said:

> Main risks I found:
> 1. WebGL requirement
> MapLibre GL JS requires WebGL. Leaflet did not. Most modern Android WebView and iOS WKWebView support it, but older devices, disabled hardware acceleration, restricted WebViews, or privacy-hardened browsers can fail or show a blank map.
> 2. URL compatibility
> Existing /map supports initial position via ?lat=...&long=..., including two pairs for bounds: src/routes/map/+page.svelte:323-325, 1242-1253.
> The MapLibre implementation in /map-next currently restores #zoom/lat/lng hashes, but I did not see equivalent ?lat=&long= support. If embedded apps use query params for initial viewport, that would break unless preserved.
> 3. Hash compatibility mostly preserved
> Existing Leaflet URLs use #zoom/lat/lon&merchant=123.
> MapLibre’s mapHash.ts intentionally mirrors that and preserves merchant / view params, with optional extra /bearing/pitch fields. Old links should parse, and new links should remain mostly backward-compatible.
> 4. No native bridge impact
> I found no postMessage, window.ReactNativeWebView, Flutter bridge, or parent iframe messaging API in the map code. So apps should not be affected through a formal JS bridge because there does not appear to be one.
> 5. DOM/CSS scraping will break
> Any integration or test that reaches into the map DOM using Leaflet-specific selectors like .leaflet-marker-pane, .leaflet-marker-cluster, or window.L will break. MapLibre uses canvas/WebGL plus .maplibregl-* controls instead.
> 6. Zoom behavior may differ
> Current /map allows maxZoom: 21; /map-next uses maxZoom: 19. If embeds deep-link to very high zooms, behavior may change.
> My recommendation before replacing /map:
> 1. Preserve ?lat=&long=, ?language=, ?boosts, #merchant=..., and #zoom/lat/lon behavior exactly.
> 2. Add a WebGL unsupported fallback message instead of leaving embeds blank.
> 3. Test in Android WebView, iOS WKWebView, Flutter webview_flutter, and React Native react-native-webview.
> 4. Document that integrations should use URL parameters only, not Leaflet DOM/classes or window.L.
> Short answer: standard iframe/WebView embeds should be okay if URL compatibility is maintained, but WebGL support and missing ?lat=&long= handling are the two main compatibility risks.
> 
> My recommendation before replacing /map:
> 1. Preserve ?lat=&long=, ?language=, ?boosts, #merchant=..., and #zoom/lat/lon behavior exactly.
> 2. Add a WebGL unsupported fallback message instead of leaving embeds blank.
> 3. Test in Android WebView, iOS WKWebView, Flutter webview_flutter, and React Native react-native-webview.
> 4. Document that integrations should use URL parameters only, not Leaflet DOM/classes or window.L.
> Short answer: standard iframe/WebView embeds should be okay if URL compatibility is maintained, but WebGL support and missing ?lat=&long= handling are the two main compatibility risks.
> 

What do you think?

We could 'test in live' for known integrations and quicky fallback if we see issues?

PR #1061, opened 2026-06-12
feat(map): simplify search UX/UI

escapedcat:
image
image
image
image

image
image

# Single search input for the map

  Replaces the map's Worldwide/Nearby toggle with one search box — no need to pick a scope
  before searching. From the "Search UX — single input" design.

  ## How search works now

  - **Empty box → what's nearby**, with a count.
  - **Type → search everywhere** by name; nearby matches still rank first. Clearing returns
  to the nearby list.
  - Toggle and the old "nothing nearby — search worldwide" prompt removed — typing already
  covers everywhere.
  - Nearby list now shows up to **250** results (was 99).
  - A failed search keeps what you typed (shows an error) instead of wiping the box.

  ## Mobile

  - Search is a **bottom sheet anchored to the screen edge** (full-width, like the merchant
  drawer and Google Maps). Drag up to expand into the full list, or tap to open without the
  keyboard popping up — so you can browse nearby first.
  - Closes via a collapse caret or the drag handle, consistent with the merchant detail
  drawer.
  - The scale bar and map credits sit just above the sheet and stay visible; the
  tile-loading indicator is never covered.

  ## Desktop

  - Same single-input behavior; the nearby count stays visible inside the open panel.
  - The search field reads more clearly as an input.

  ## Map info

  - Nearby places/count now appear immediately on load (previously only after panning or
  opening the panel).
  - Map credits collapse to a small (i) on mobile to keep the bottom clear; full credits
  remain on desktop.

  ## Worth confirming (intended, not bugs)

  - 1–2 characters keeps the nearby list; worldwide search starts at 3 characters.
  - Collapsing the mobile sheet clears the active search (re-opening starts fresh on
  nearby).

  ## Analytics

  - Retire: `worldwide_mode_click`, `nearby_mode_click`, `search_worldwide_cta_click`.
  - New mobile-sheet events: `search_sheet_tap_expand`, `search_sheet_swipe_expand`,
  `search_sheet_swipe_collapse`.

  Key change from your draft: the mobile section now describes the anchored bottom sheet
  (the latest direction), not a floating card.


dadofsambonzuki:
Very nice 😍

dadofsambonzuki:
If I seqch for a place then select it then swipe back I get this

escapedcat:
@dadofsambonzuki wanna give it one last try?

dadofsambonzuki:
All good 🤟

PR #1062, opened 2026-06-14
fix(merchant): preserve line breaks in comments #662

### LINKED ISSUE #662: feat: improve multi-paragraph comment handling
Comments should respect line-breaks

See i.e.:  
https://btcmap.org/merchant/34519#comments

### Screenshots

Image

Image

b0rgbart3:
Fixes: https://github.com/teambtcmap/btcmap.org/issues/662

Change made:
Render merchant comment text with white-space: pre-line so line breaks entered in the comment textarea are preserved when displayed, instead of being collapsed by the browser's default HTML whitespace rules.

Before:
before662

After:
after662

## Summary by CodeRabbit

* **Style**
  * Improved comment text display to properly preserve line breaks and whitespace formatting.

### DIFF
--- a/src/routes/merchant/[id]/components/MerchantComment.svelte
+++ b/src/routes/merchant/[id]/components/MerchantComment.svelte
@@ -35,7 +35,7 @@ export let compact = false;
 	>
 		<div class:space-y-2={!compact} class:space-y-1={compact} class:lg:space-y-0={!compact}>
 			<span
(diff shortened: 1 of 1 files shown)


escapedcat:
Instead of

> This PR addresses Issue https://github.com/teambtcmap/btcmap.org/issues/662

use

> Fixes: #662

In that case github will reference it by itself

Issue #160, opened 2024-10-14
Add heatmap layer to the map


dadofsambonzuki:

escapedcat:
https://maplibre.org/maplibre-gl-js/docs/examples/create-a-heatmap-layer/

dadofsambonzuki:
Tackling this via #1082

PR #1082, opened 2026-06-23
feat(map): add merchant heatmap layer #160

dadofsambonzuki:
Add a toggleable merchant-density heatmap layer to the MapLibre merchant map.

- feed the heatmap from a dedicated non-clustered GeoJSON source so it reflects overall place density
- persist the toggle state and add localized labels for the control

## Summary by CodeRabbit

* **New Features**
  * Added a merchant-density heatmap layer to the map with a flame-style toggle, including persisted enabled state and consistent behavior across basemap changes.
  * Heatmap visibility now adapts to zoom to prevent pins/clusters from visually competing when zoomed out.
* **Localization**
  * Added translated label and accessible text for the heatmap toggle in all supported languages.
* **Analytics / Tracking**
  * Heatmap toggle interactions now emit an analytics event with the enabled/disabled state.
* **UI Polish**
  * Improved toggle enabled-state icon styling, including a dark mode adjustment.
* **Tests**
  * Expanded end-to-end coverage for rendering, state toggling, persistence across reloads, and console error checks.


escapedcat:
Might do a follow up to move certain options to a settings modal

image

escapedcat:
image

Looks like it's possible to pay "everywhere" in i.e. Europe and USA :D
Any way to adjust this a bit? Or will it turn purple when there's more in one place? Do we need a legend? Should this be it's own view with a locked zoom because on low zooms this looks "spotty"? 
image
Who's the audience of such view? Daily users? Is it a marketing/promotional view? More like for people that also check the analytics?

Just random thoughts really.

dadofsambonzuki:
Have adjusted the settings, looks better now.

image

Don't think we need a legend for now. Maybe later when more merchants and the color differences matter more.

Low zooms being spotty not a big issue imo as it depends where you are. Berlin El Sal looking good!

image

Audience? Well, I like it 🙃 But yes, also good for marketing & socials. Sort of thing people pick up on and reference.

escapedcat (approved):

PR #1083, opened 2026-06-23
fix: heatmap review followups

escapedcat:
tackling some review issues


dadofsambonzuki:
I think i addressed some of these on the original PR. Can you not make changes there?

I'll stop on that other PR now.

PR #990, opened 2026-05-10
feat(account): auth nostr #903

escapedcat:
Partly fixes: #903

image

## Summary by CodeRabbit

* **New Features**
  * Added Nostr authentication on the login page, supporting sign-in via a Nostr browser extension or by entering an **nsec** secret key.
  * Added server-backed Nostr session creation/validation and extended the session to optionally store an associated **Nostr public identity**.
  * Updated login UI copy with localized Nostr/nsec instructions, progress, and error messages.
* **Bug Fixes**
  * Improved logout to clear the session and redirect to the login page, avoiding inconsistent post-logout UI state.
* **Tests**
  * Expanded Nostr and session coverage for new authentication and `npub` handling.

PR #1086, opened 2026-06-24
feat(nostr): show Nostr profile avatar in UserMenu + test backfill

escapedcat:
## What

Follow-up to #990. When the logged-in account was authenticated via Nostr (`session.npub` set), show the user's Nostr profile avatar in the header `UserMenu`, falling back to the generic account icon otherwise.

Also backfills tests for the merged Nostr login code (in progress — see commits).

## Commits so far
- `nostrProfile.ts` — fetch kind:0 metadata via `SimplePool.get` (3s max-wait, SSR-guarded, session-cached, never rejects).
- `nostrProfile.test.ts` — unit tests (parse / cache / failure modes).
- `NostrAvatar.svelte` — thin component; `` with icon fallback on loading / no-profile / image error.
- Wire-in to `UserMenu` trigger.

## Still to come (this PR)
- Proxy handler unit test (`/api/session/nostr`).
- Extract + test `parseNostrAuthResponse` from `NostrLoginForm`.

## Testing the avatar
Log in with Nostr using an **nsec whose pubkey has a published kind:0 profile** to see the picture; a freshly-generated key has no profile, so the avatar correctly falls back to the account icon.

Draft until the test backfill commits land and it's reviewed.

## Summary by CodeRabbit

* **New Features**
  * Added Nostr profile avatars across the app, loading the user’s picture automatically with graceful fallback while loading or on failure.
  * Enhanced Nostr account display to show a fetched profile name when available (instead of the raw username).

* **Bug Fixes**
  * Improved Nostr login/session handling by validating and parsing session responses more reliably.
  * Prevented stale profile data from briefly showing during rapid account changes (avatar/name updates now correspond to the latest identity).


src/lib/nostrProfile.ts
```diff
@@ -0,0 +1,109 @@
...
+	picture?: string;
+	nip05?: string;
+};
+
+// Small hardcoded starter relay set queried for kind:0 metadata; first relay
+// to answer wins. NIP-65 (kind:10002) relay discovery is a later refinement.
+const RELAYS = [
+	"wss://relay.damus.io",
+	"wss://nos.lol",
+	"wss://relay.nostr.band",
```
dadofsambonzuki: nostr.band is ded, but the LLMs still love it! I would add in primal and ditto for now.

src/lib/nostrProfile.ts
```diff
@@ -0,0 +1,109 @@
...
+// a profile may set any subset (or publish nothing at all).
+export type NostrProfile = {
+	name?: string;
+	displayName?: string;
+	picture?: string;
+	nip05?: string;
+};
+
+// Small hardcoded starter relay set queried for kind:0 metadata; first relay
+// to answer wins. NIP-65 (kind:10002) relay discovery is a later refinement.
```
dadofsambonzuki, line 14: Came here to say we should use this, but fine to do it later under a relay management PR. Usually best to have a core set plus the NIP-65 set. We may want to use relay.btcmap.org to blast to other public relays and save the client making multiple connections.

dadofsambonzuki (changes requested):

src/lib/nostrProfile.ts
escapedcat: Added #1088

dadofsambonzuki (approved):

PR #1094, opened 2026-06-25
fix(boost): clear, retryable error when invoice generation fails #1092

### LINKED ISSUE #1092: Cannot boost merchants
When I click on "Boost" in the node on a desktop computer and then choose "Boost for 1 month", nothing happens. I only see a map with the various merchants without any navigation options. If I click on a merchant, I don't get to see the business details. The only option is to close the app. 

Mobile reacts the same way.

Image

Image

escapedcat:
## Problem

When boosting a place from the map merchant drawer, a failed invoice request (`POST /api/boost/invoice/generate`) only flashed a transient, English-only toast and then left the drawer on the duration-selection screen unchanged. Users couldn't tell what happened or what to do next — and the boost endpoint is currently failing upstream, so this path is hit in practice.

## Change

Replace the silent toast with a dedicated, retryable error step inside the boost drawer:

- **Dedicated error step** — warning icon, plain-language cause, a **Try again** button (retries with the already-selected duration), and a `support@btcmap.org` contact link.
- **Cause-specific messaging** via a small pure helper `classifyBoostError`: *network* ("You appear to be offline…") when no response comes back vs *service* ("The boost service is temporarily unavailable…") when the server answered with an error. Raw HTTP status codes are never shown.
- **Localization** — the two previously hardcoded-English error strings (`PAYMENT_ERROR_MESSAGE`, `STATUS_CHECK_ERROR_MESSAGE`) moved into the i18n `errors.*` namespace (also covering the comment flow and QR-load toasts), and all new strings translated across the 8 non-English locales.

Scope is intentionally limited to the invoice-generation failure; the in-payment polling and post-payment finalize toasts are unchanged.

## Testing

- `pnpm run format:fix`, `pnpm run check` (0 errors), `pnpm run lint` (clean), `pnpm run test --run` (441 pass).
- New unit tests for `classifyBoostError` cover both the network and service branches.
- Verified live via DevTools that the click → `generateInvoice` → error-handling path executes; deliberately did **not** add an e2e test that hits the production API.

Closes #1092

---

## Update: root-cause fix for the intermittent "boost UI disappears" report

While testing, the boost drawer would intermittently reset to a blank stage 0 (no invoice, no selected duration) instead of showing the error step. Root cause was **not** in the boost UX code — it's the **service worker** (`src/service-worker.ts`):

- It intercepted every GET, including `api.btcmap.org` place data and `/api/boost/invoice/status`, and cached them.
- On any network hiccup it served a **stale cached** response or the **`offline.html`** fallback.
- Returning HTML where the app expects JSON corrupted the merchant lookup → `merchant.id` became invalid → `generateInvoice()` hit its `Invalid merchant ID` guard (or `merchant` went null and `BoostContent` unmounted), so the drawer silently reset. Intermittent because it depended on cache/network timing.

Confirmed by the SW cache literally holding hundreds of `api.btcmap.org` entries, and an API GET returning the app-shell HTML when the SW was active.

The new commit makes the SW **bypass API/data requests** (`api.btcmap.org`, `/api/`, `/rpc`, `/btcmap-api-proxy`) so they always hit the network and return real JSON or a real error. Verified: after the change the SW caches **0** API responses. Trade-off noted in the commit — API data is no longer served from the SW cache offline (the app surfaces a network error instead of stale/corrupted data).

## Summary by CodeRabbit

* **New Features**
  * Boost flow now shows an in-page, localized error panel (network vs. service) with a retry option when invoice generation or status checks fail.
  * Added/expanded localized translations for boost and invoice-related error states across supported languages.
* **Bug Fixes**
  * Invoice/boost error handling now uses i18n-driven messages instead of removed constant-based text.
  * Service worker no longer serves cached/offline responses for API/data-like routes.
  * Improved payment QR confetti layering when the invoice is marked as paid.
* **Tests**
  * Added unit tests for boost error categorization and invoice “paid” detection.


escapedcat:
@dadofsambonzuki couldn't really test this. Getting blocked a lot.

dadofsambonzuki:
tested and lgtm

dadofsambonzuki:
all good functionally, will leave you to clear the comments when you have time.

dadofsambonzuki (approved):

Issue #921, opened 2026-04-16
idea: unified issues


escapedcat:
- ultimate goal should be unified issues (linting, add locations, verify, community, import) within our own platform with collaboration (assignment, commenting, etc.)
- The correct approach is
  1. Make areas discoverable on the map
  2. Leveraging that, make area issues discoverable and actionable, that's the only way to involve the locals. Providing free service for people who can't bother creating an osm account doesn't help with anything. They send their low effort submissions and we get a new outdated location in a year because they don't even know how to bump check dates since they weren't the ones adding places. The end result is degraded user experience

escapedcat:
## Feasibility spike: the `?issues` map (2026-08-05)

Ran the API/architecture spike for the "make issues discoverable on the map" step of this idea (the follow-up to `/map?outdated`, per the community feedback: issues map + GPS locate, fix-hint popups, complexity filters). **Verdict: feasible now, and cheaper than expected — the long-standing blocker dissolved.**

### What the API offers (probed live)

- `/v4/place-issues` **requires `area_id`** — there is no world query. The `earth` area (id 662, type `planet`) exists but carries only **50 issues, all `missing_icon`** — apparently just the places outside every country. Worldwide coverage via this endpoint would mean ~250 per-country requests: not viable client-side.
- Issue rows carry **only OSM identity** (`element_osm_type`/`element_osm_id`), no numeric place id — the osm-id join gap is real *if* the API were the source.
- `/v4/places?fields=id,osm_id` (the enrichment that would close the join) measures **1.18 MB raw** — same scale as the already-shipped `id,verified_at` enrichment (1.13 MB), so it's viable *when needed*. It isn't needed for the MVP:

### The blocker dissolves: issues are client-derivable

The real-world issue distribution (US: 1318 issues, DE: 478) is:

| code | US | DE | derivable from |
|---|---|---|---|
| `outdated` | 940 | 397 | `verified_at` (already enriched for `?outdated`) |
| `not_verified` | 258 | 28 | `verified_at` absent |
| `outdated_soon` | 96 | 51 | `verified_at` math |
| `missing_icon` | 24 | 2 | `icon` field (in the bulk feed) |
| tag-lint types | **0** | **0** | server-only — and absent from every sampled set |

**100% of the sampled issue volume derives from data the client already has.** The `verified_at` enrichment shipped with the verified-recency filter; `icon` is in the bulk feed. No new payloads, no osm join, no API changes.

### MVP sketch

[...]

PR #1155, opened 2026-07-30
fix(map): restore the ?outdated URL query filter #149

escapedcat:
## What

Restores the `/map?outdated` deep link: a re-verification worklist view showing only places whose last verification is missing or older than 12 months. Also surfaces the mode as a fifth **Outdated only** option in the Layers & filters modal so it's discoverable without knowing the param.

### Testing / Screenshots

https://deploy-preview-1155--btcmap.netlify.app/map?outdated

image

## Why / history

- `?outdated` worked from 2024-04 (0ba6c7fc) until the cold-start rewrite (#307, 2025-09) silently dropped it — the v4 bulk feed carries no verification dates, so the classification vanished with the old marker loop. `?boosts` broke in the same commit and was restored in #1097; `?outdated` never was.
- The verified-recency filter (#1081) has since solved the data gap (`ensureVerifiedDates()` lazily merges `verified_at` into `$places`), so the deep link now rides that machinery instead of a parallel mechanism: `"outdated"` is a new inverse mode of `VerifiedFilterYears`. Every existing chokepoint (markers, nearby list, radius-API list, search results) applies it for free, and "verified within N years" vs "outdated only" are mutually exclusive by construction — no empty-map intersection.

## Behavior

- `?outdated` (bare key, any value — matching the legacy form) seeds the filter for the session **without persisting**, so a shared link can't overwrite a visitor's stored preference. Picking **Outdated only** in the modal persists like any other option.
- Switching the filter away strips the stale param via `history.replaceState` (mapHash idiom) so a reload doesn't resurrect it.
- The filter stays inert until the one-time `verified_at` enrichment lands (`verifiedDatesLoaded` gate) — same as the existing recency filter, avoiding a flash of "everything is outdated".
- Never-verified places count as outdated, matching legacy semantics and the Android classification.

## Test plan

- Unit: new coverage for the inverse predicate (`filterPlacesByRecency`) and storage round-trip incl. unknown-value rejection; full suite 466 passing.
- Browser (dev server): `/map?outdated` narrows to 12,143 pins vs 12,134 ground-truth outdated from `/v4/places?fields=id,verified_at` (9-place drift = fetch timing); modal shows Outdated only checked + accent dot; deep link with a stored `within 1 year` preference does not overwrite it; selecting Any strips the param and restores the full set; cold visitor auto-triggers enrichment.
- `format:fix` / `check` / `lint` all clean.

## Notes for review

- i18n: `verificationFilter.outdatedOnly` added to all nine locales, phrased after each language's existing `boostedOnly` label.
- Pre-existing quirk noticed (not addressed here): the `$lastUpdatedPlaceId` re-sync path pushes unfiltered `$places` to the map source, momentarily bypassing every filter (verified/category/boost) until the next reactive rebuild. Worth a follow-up if it bites.

Refs #149 (map view for tagging issues — this restores the outdated slice of it) · context: #921 (unified issues)

## Summary by CodeRabbit

- **New Features**
  - Added an “Outdated only” verification filter to show locations that have not been verified recently or have no verification date.
  - Available across all supported languages.
  - Added an `?outdated` map link for quickly opening this filtered view.
- **Improvements**
  - The outdated filter applies for the current session without replacing the saved filter preference.
  - Filter selections now persist and restore reliably, including the new outdated option.


dadofsambonzuki:
Perhaps for a follow-up PR, but the list view should perhaps be issues? Or perhaps a Places list with issues count and an issues centric detail pane? 'Fix on OSM' etc.

escapedcat:
Addressed the AI review feedback in the latest push:

- **MapControls importing merchantListStore** (Copilot + qodo): adopted the prop approach — the page passes `currentVerified={$merchantList.verifiedWithinYears}` and the component's local `selectedVerified` state is gone entirely (the store update in `applyVerifiedFilter` is synchronous, so the prop drives the radio + accent dot with no mirrored state). `/communities/map` no longer touches the store. (abc97cb3)
- **`filterPlacesByRecency` accepting any `number`** (Copilot): narrowed. The `1 | 2 | 3 | "outdated" | null` union now lives in `verification.ts` as `RecencyFilter` (keeping the generic lib free of map-UI imports) and `VerifiedFilterYears` aliases it — one source of truth, callers unchanged. (3dcee17b)

Also added e2e coverage (b7e195e5) so the param can't be silently lost in a future rewrite like #307: count-narrowing vs unfiltered baseline, no-persist contract, modal checked state, and param-stripping.

escapedcat:
> Perhaps for a follow-up PR, but the list view should perhaps be issues? Or perhaps a Places list with issues count and an issues centric detail pane? 'Fix on OSM' etc.

This is just re-enabling the lost functionality. Further improvements/features regarding issue maintenance will be follow-up. [grill-with-docs](https://www.aihero.dev/skills-grill-with-docs) is on it.

dadofsambonzuki (approved):

Issue #1168, opened 2026-08-01
refactor(map): visiblePlaces — one pipeline for which places are visible


escapedcat:
**Recommendation: Strong — top pick of the review.**

## Problem
"Which places are visible" is decided at **10 sites with 3 divergent readiness gates**: the marker reactive block (`+page.svelte:890-936`), `updateMerchantList` (`:399-409`), `syncPlacesToSource` (`:821-829`), `fitSearchResultBounds` (`:526-531`), the `$lastUpdatedPlaceId` block (`:944`), `setMerchants` / `fetchAndReplaceList` / `fetchCountOnly` / `openWithSearchResults` (`merchantListStore.ts:201,273,329,417`), and the panel's `filteredSearchResults` (`MerchantListPanel.svelte:357`).

Every one of these live bugs is a pair of those sites disagreeing: #1158 (sticky unfiltered pins), #1159 (category not on pins in search), #1160 (camera fits hidden results), #1161 (unfiltered count badge), #1162 (stale category counts).

## Sketch
```ts
// src/lib/map/visiblePlaces.ts
selectVisiblePlaces(inputs: PlaceFilterInputs): Place[]   // pure: search/category/recency+gate/boost/deleted
export const visiblePlaces: Readable<Place[]>             // derived(places, merchantList, verifiedDatesLoaded)
export const visibleSignature: Readable<string>           // replaces the 5 hand-kept memo vars (+page.svelte:929-933)
```
Every consumer (pins, list, counts, camera) reads the store; none re-filters. `$places` goes out of scope at render sites, making the #1158 bug class unrepresentable.

## Wins
- Locality: one file answers "why is this pin/row here?"; the readiness gate exists once, not three times
- Leverage: the planned `?issues` map filter (#921) lands in one function instead of 10 sites — `?outdated` (#1155) just had to thread through 4 of them
- Testability: the pipeline becomes a pure function beside `verifiedFilter.test.ts`; today its divergences are only reachable through WebGL Playwright

Fixes the root cause of #1158, #1159, #1160, #1162; pairs with the nearbyList re-cut for #1161.

---
_Found by a Claude Code architecture review (2026-08-01, map-view focus)._

escapedcat:
**Re-review verdict: RE-SCOPE before implementation.**

[...]

Issue #1162, opened 2026-08-01
bug(map): search-mode category counts go stale after changing the verified filter


escapedcat:
## Problem
In search mode, changing the verified filter re-filters the rendered list but **not** the category counts — chips' enabled/disabled state and the "N of M" line disagree with what's shown.

## Evidence
- Counts frozen at search time: `src/lib/merchantListStore.ts:424-427` (`openWithSearchResults`)
- List re-derives reactively: `src/routes/map/components/MerchantListPanel.svelte:357-362`
- `applyVerifiedFilter` ends in `updateMerchantList({force:true})`, which is a no-op in search mode: `src/routes/map/+page.svelte:383` (early return)

## Fix direction
Make `categoryCounts` a derived value of the same filtered selection the list renders, instead of a snapshot. See linked visiblePlaces refactor issue.

---
_Found by a Claude Code architecture review (2026-08-01, map-view focus)._

PR #1177, opened 2026-08-01
fix(map): stop the single-place resync from bypassing every filter #1158

### LINKED ISSUE #1158: bug(map): filtered pins reappear and stick after boosting/commenting while a filter is active
## Problem
Boosting or commenting on a merchant while any map filter is active (verified/outdated, category, search, boosts) brings every hidden pin back — and they **stay** back until an unrelated event (locale change, save toggle, filter change) forces a rebuild.

## Evidence
`src/routes/map/+page.svelte:944`:
```js
$: if (map && styleLoaded && $lastUpdatedPlaceId) {
	syncPlacesToSource($places);   // unfiltered — bypasses search/category/recency
	lastUpdatedPlaceId.set(undefined);
}
```
Two defects compound:
1. It passes raw `$places`, skipping every filter except `boostsOnly` (which lives inside `syncPlacesToSource`).
2. It updates none of the memo variables (`lastPlacesLength` etc., `:929-933`), so the next run of the marker reactive block compares the filtered count against a stale memo, finds no change, and **skips the corrective re-sync**. The unfiltered map is sticky.

## Repro
1. Open /map, set "Verified within 1 year" (or run a search)
2. Open a merchant drawer, boost it (or add a comment) — `BoostContent.svelte:61` fires `lastUpdatedPlaceId`
3. All hidden pins reappear and persist

## Fix direction
Short-term: apply the same filter pipeline in this block and refresh the memo signature. Root cause: a single visible-places seam (see the linked refactor issue) makes this class of bug unrepresentable — `$places` should not be in scope at any render site.

---

escapedcat:
Fixes #1158.

## What was happening
Boosting or commenting on a merchant fires `lastUpdatedPlaceId`, whose reactive block synced **raw `$places`** to the map source — bypassing search/category/recency filtering (in search mode even the boosts filter, so it pushed all ~29k places). Worse, it updated none of the marker block's memo variables, so the next reactive run compared the filtered count against a stale memo, saw "no change", and skipped the corrective resync — **the unfiltered map was sticky**. The direct sync also poisoned `lastSyncedList`, which the heatmap seed and the boosted-cluster zoom-crossing resync redraw.

## The fix
One line: invalidate the memo (`lastPlacesLength = -1` — the same idiom the style-load handler already uses) instead of syncing directly. The marker block re-runs in the same flush, recomputes `effective` with all current filters, refreshes every memo coherently, and resyncs. This also un-poisons `lastSyncedList` for the heatmap/zoom-crossing paths for free.

Deliberately **not** extracting shared filter logic here — that is the #1168 `visiblePlaces` refactor; this is the minimal correct fix.

## Verification
Boost/comment both POST to prod, so no e2e (per project test policy). Verified in the browser with a temporary dev-only hook calling `updateSinglePlace(id)` (GET) + `lastUpdatedPlaceId.set(id)` — the exact production trigger path:
- un-fixed: filtered 16,159 pins → jumped to 29,054 and **stayed** across waits (bug reproduced)
- fixed: stays 16,159 immediately and after settling

`format:fix` / `check` / `lint` / 466 unit tests clean.

## Known trade
In search mode a just-boosted pin no longer force-repaints its boost ring instantly (search results are store-held objects, not $places-sourced). The old code only "achieved" that repaint by destroying the filtered view; keeping the view correct wins.

## Merge order
Land this one first among #1158-#1162 — smallest diff; siblings rebase trivially. Whoever does #1168 must preserve or replace the memo-invalidation contract.

### DIFF
--- a/src/routes/map/+page.svelte
+++ b/src/routes/map/+page.svelte
@@ -938,11 +938,13 @@ $: if (map && styleLoaded && $places) {
 // In-place mutations to a place (boost confirmation, new comment count)
 // don't change array length or any of the size counters above, so the
 // guard wouldn't repaint the pin. lastUpdatedPlaceId is set by
-// updateSinglePlace() in $lib/sync/places.ts — re-sync the source when
-// it fires, then clear so we don't re-sync next time anything else
-// triggers this reactive block.
+// updateSinglePlace() in $lib/sync/places.ts — invalidate the memo so the
+// marker block above re-runs, recomputes `effective` with the current
+// search/category/recency filters applied, and re-syncs (same idiom as the
+// style-load reset). Never sync raw $places here: that bypasses every
+// filter and poisons lastSyncedList for the heatmap/zoom-crossing paths.
 $: if (map && styleLoaded && $lastUpdatedPlaceId) {
-	syncPlacesToSource($places);
+	lastPlacesLength = -1;
 	lastUpdatedPlaceId.set(undefined);
 }
 
--- a/src/routes/map/+page.svelte
+++ b/src/routes/map/+page.svelte
@@ -882,7 +882,8 @@ const triggerEnrichmentIfNeeded = debounce(() => {
 
 // In-place mutations to a place (boost confirmation, new comment count)
 // don't change array length or any of the size counters below, so the marker
-// block's guard wouldn't repaint the pin. lastUpdatedPlaceId is set by
+// block's guard wouldn't repaint the pin. lastUpdatedPlaceId is set by the
+// boost/comment flows (BoostContent, CommentAdd) right after they await
 // updateSinglePlace() in $lib/sync/places.ts — invalidate the memo so the
 // marker block re-runs with the current search/category/recency filters and
 // re-syncs (same idiom as the style-load reset). Never sync raw $places here:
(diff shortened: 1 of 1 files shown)


escapedcat:
Adversarial re-review (7-agent pass over all six PRs) found a **confirmed high-severity flaw in the first version of this fix**, now corrected in the latest push:

The memo invalidation never actually re-ran the marker block. Svelte 4's compiler excludes self-assigned variables when topologically ordering reactive blocks (`add_declaration` skips `assignees`), so the marker block — which also assigns `lastPlacesLength` — was not ordered after the invalidation block, and a mid-update `$$invalidate` cannot re-queue the component (`make_dirty` requires idle). Net effect: #1158's bypass was gone, but the repaint this block exists for silently stopped firing, in all modes. The original browser QA measured only pin **count**, which stays correct trivially once the bad sync is removed — it structurally could not detect the lost repaint.

**Fix:** the invalidation block now sits **above** the marker block — within one `$$.update()` pass, later blocks do see dirty bits OR'd by earlier ones, and nothing re-sorts the pair (the marker block assigns none of the invalidation block's dependencies). The comment states the ordering contract explicitly for #1168.

**Re-verified with an instrumented `syncPlacesToSource` call counter** (closing the QA gap): the simulated boost trigger now increments the counter (repaint fires) while the count stays at the filtered 16,160 — both properties, not just one.

Remaining known trade, unchanged: search-mode pins render from `searchResults`, which `updateSinglePlace` doesn't refresh, so a searched pin's boost ring repaints only data the store already holds — pre-existing class, noted in the comment.

escapedcat:
Both Copilot findings adopted (e148cc0e): the guard is now an explicit `!== undefined` check (the truthy form would skip a place id of 0), and the comment no longer credits `updateSinglePlace()` with setting `lastUpdatedPlaceId` — the boost/comment flows set it after awaiting that refresh. The wording was carried over from main, but wrong is wrong.

Issue #1160, opened 2026-08-01
bug(map): 'Show all on map' fits bounds of results the user cannot see


escapedcat:
## Problem
The "Show all on map" button label/disabled state derives from the **filtered** search results, but the camera fits the bounds of the **unfiltered** ones — with a category or verified filter active it frames places the user cannot see.

## Evidence
- Button state: `src/routes/map/components/MerchantListPanel.svelte:672-679` (uses `filteredSearchResults`)
- Camera: `src/routes/map/+page.svelte:526-531` `fitSearchResultBounds` reads `get(merchantList).searchResults` (raw)

## Fix direction
Fit the same list the button counts. Falls out for free from a single visible-places selection (see linked refactor issue).

---
_Found by a Claude Code architecture review (2026-08-01, map-view focus)._

PR #1178, opened 2026-08-01
fix(map): apply the category filter to search-mode pins #1159

### LINKED ISSUE #1159: bug(map): category filter does not apply to map pins in search mode
## Problem
In search mode the category chips filter the merchant **list** but not the map **pins** — the map keeps showing every search result.

## Evidence
- Marker reactive block takes the search branch before the category branch: `src/routes/map/+page.svelte:895-899`
- The panel filters for the list separately: `src/routes/map/components/MerchantListPanel.svelte:357-362`
- `modeSig` (`+page.svelte:916-920`) omits the category in search mode, so even adding the branch wouldn't re-sync without also fixing the signature.

## Fix direction
Apply category filtering to the search-mode pin set (and include it in the change-detection signature) — or route both list and pins through one shared visible-places selection (see linked refactor issue).

---

escapedcat:
Fixes #1159.

## What was happening
In search mode the marker pipeline's search branch rendered **raw `searchResults`** — the category chips filtered the merchant list (via the panel's `filteredSearchResults`) while the map kept every pin. The change-detection signature also omitted the category in search mode, so even a same-length category switch never re-synced.

## The fix
Two small edits in the marker reactive block:
- the search branch now filters with `placeMatchesCategory` — deliberately the **panel's exact predicate**, so list and pins agree by construction (the repo's two category predicates are only extensionally equal today; tying pins to the other one would recreate this bug class the day they diverge)
- the search signature includes the category (`s:${category}:${ids}`), so chip changes re-sync even at equal counts

Recency already reached search pins (the `inSearch` gate bypass); category was the only gap. Filter order matches the panel (category → recency).

## Verification
- Live (real GET-only search): "coffee" → 100 results/100 pins; Coffee chip → panel "82 of 100 result(s)", pins **82**; All → 100. Pre-fix pins stay at 100.
- New e2e (`tests/map-search-category-filter.spec.ts`): stubbed search route, chip narrows `__mapPlacesCount` 3→2→3 — fails on pre-fix code by construction. GET-only, no prod writes.
- `format:fix` / `check` / `lint` / unit suite clean.

## Notes
- Side effect (consistent, intentional): with the heatmap overlay on, search-mode density now also respects the chip.
- Known interim: "Show all on map" still fits **unfiltered** results until #1160 lands — this fix makes that mismatch more visible, sibling PR addresses it.
- Latent footgun documented in the commit: `filterMerchantsByCategory` vs `placeMatchesCategory` consolidation belongs to #1168.

## Merge order
Independent of #1177 (#1158) — different regions of the same file; either order.

### DIFF
--- a/src/routes/map/+page.svelte
+++ b/src/routes/map/+page.svelte
@@ -158,10 +158,12 @@ let lastSavedIdsSize = -1;
 let lastEnrichedCacheSize = -1;
 let lastLocale: string | null | undefined;
 let lastAppliedLabelTheme: "light" | "dark" | undefined;
-// Signature of the search-mode visible set. Empty string forces an
-// initial sync; "n" = nearby (all places); "s:<id>,<id>,…" = a specific
-// search-result list. Distinct from lastPlacesLength because a fresh
-// search may have the same result count as the previous one.
+// Signature of the mode-dependent visible set. Empty string forces an
+// initial sync; "n" = nearby (all places); "c:<category>" = nearby
+// narrowed by a chip; "s:<category>:<id>,<id>,…" = a search-result list
+// (optionally narrowed by a chip); the sync guard appends "|v:<years>"
+// for the verified-recency window. Distinct from lastPlacesLength
+// because a fresh search or chip switch may leave the count unchanged.
 let lastSearchModeSig = "";
 
 // Last place list fed to the sources, kept so the boosted-clustering boundary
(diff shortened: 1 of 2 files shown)


escapedcat:
qodo feedback dispositions:

1. **Stale modeSig comment** — adopted (0809d53f): the `lastSearchModeSig` declaration comment now documents the full current grammar (`n` / `c:<category>` / `s:<category>:<ids>` plus the `|v:<years>` suffix the guard appends).
2. **e2e uses the live places API** — declined for this PR: `waitForMarkersToLoad()` against the live API is the established pattern of every map spec in this suite (map-tools-panel, heatmap-toggle, merchant-list-panel, map-outdated-filter); making only this spec hermetic would diverge from the suite convention while leaving the same exposure everywhere else. A suite-wide hermetic-fixtures pass is a reasonable follow-up if the team wants it — happy to file it as an issue.

Issue #1171, opened 2026-08-01
refactor(map): nearbyList — one radius request, one list status


escapedcat:
**Recommendation: Worth exploring.**

## Problem
Three byte-identical radius requests with three different filter policies (`merchantListStore.ts:253,340,382`), and the list's tri-state ("zoom in" / too-dense / truncated) is decided in the store (`:281-293`), requested by the page (`+page.svelte:436`), and re-derived by the panel from residue (`MerchantListPanel.svelte:395-398`). `fetchCountOnly` skips every filter — that's #1161. Deletion-test findings: `fetchCountOnly` and `setMode` are pass-throughs; `fetchEnrichedDetails` is ~6 real lines wrapped in 33.

## Sketch
```ts
// one private searchPlacesInRadius(center, radiusKm, fields, signal) + three reducers
nearbyList.refresh({ center, bounds, zoom, force })
nearbyList.status: Readable<'below-floor'|'too-dense'|'truncated'|'empty'|'ok'>
```

## Wins
- Locality: `MERCHANT_LIST_FETCH_CEILING` / `MERCHANT_LIST_LOW_ZOOM` stop living in three modules
- Fixes #1161 at the root: the badge counts the same filtered set the list shows

---
_Found by a Claude Code architecture review (2026-08-01, map-view focus)._

escapedcat:
**Re-review verdict: RE-SCOPE, and demote below #1172 (swap the order).**

**Stale evidence (motivation half-consumed by merged fixes):**
- "Three byte-identical radius requests" — no longer true: #1180 gave `fetchCountOnly` a distinct conditional-fields request; only `fetchAndReplaceList` and `fetchEnrichedDetails` remain identical.
- "`fetchCountOnly` skips every filter — that's #1161" — fixed by #1180 (snapshot, recency-filtered count, stale-response guard); category exclusion is now a documented invariant with tests.
- Still true: the tri-state smear (store blanks/keeps, page owns the ceiling, panel reconstructs "too dense" from residue) and the `setMode` pass-through.

**Design flaws in the status-enum sketch:**
- `below-floor` cannot derive from zoom alone: `?boosts=true` forces local-markers at ANY zoom — `refresh()` needs the behavior override, not just `{center, bounds, zoom}`.
- Timing regression: the panel's zoom-in prompt reacts to live zoom per moveend; a status computed only at debounced refresh time lags pinch-zooms 300ms+ and goes stale in search-mode pans.
[...]

PR #1180, opened 2026-08-01
fix(map): count the filtered set in the closed-panel badge #1161

### LINKED ISSUE #1161: bug(map): nearby count badge ignores all filters at zoom 10-14 while the panel is closed
## Problem
At zoom 10-14 with the panel closed, the nearby count pill / search-bar badge shows the **unfiltered** place count — verified/outdated filter, category, and boosts are all ignored.

## Evidence
`src/lib/merchantListStore.ts:329-365` `fetchCountOnly` requests `fields=id` and writes `totalCount` straight from the raw id array — no recency, no category, no boost. That `totalCount` feeds `NearbyCountPill` (`MerchantListPanel.svelte:299,568-570`) and the search-bar `nearbyCount` (`+page.svelte:2093`).

## Fix direction
`fetchCountOnly` duplicates the radius request byte-for-byte with different fields and skips the filter policy; it fails the deletion test — fold it into the main radius fetch and count the filtered set (fetching `id,verified_at,icon` costs little). See linked nearbyList refactor issue.

---

escapedcat:
Fixes #1161.

## What was happening
At zoom 10-14 with the panel closed, `fetchCountOnly` requested `fields=id` and wrote `totalCount` straight from the raw array — the nearby pill and search-bar badge ignored the verified/outdated filter. This fired on **first paint** for `?outdated` deep links and returning users with a persisted window: filtered markers, unfiltered count.

## The fix
Store-local, inside `fetchCountOnly`:
- snapshot the filter once so the `fields` param and the post-response filter can never disagree (a mid-flight change re-invokes and aborts)
- widen to `fields=id,verified_at` **only while a filter is active** — the common filter-off request stays byte-identical to today
- count the recency-filtered set

**Deliberately recency-only** (analysis findings): category would be dead code — `close()` resets it to `'all'` and this path only runs with the panel closed; boost needs nothing — `?boosts=true` forces the local-markers path, making this fetch unreachable. The policy exactly matches `fetchAndReplaceList`'s `hideIfExceeds` branch (recency-only), the state users transition into on open — so the count no longer jumps at the open/close boundary. A filtered count of 0 now correctly falls through to "no merchants" instead of a phantom "zoom in" prompt.

## Verification
- Unit (TDD, red→green): fields stay `id` with filter off (tightened to assert **absence** of `verified_at`), widen with filter on, year-window counting, outdated-mode counting (stale + never-verified). 68 store tests green.
- Live at zoom 12 over Hamburg: badge **18 nearby** unfiltered vs **7 nearby** with `?outdated` (pre-fix: 18 both).

## Notes
Pre-existing and intentionally untouched: closing a panel that had a category selected leaves the category-filtered count on the pill until the next moveend — #1171's territory.

## Merge order
Independent of the sibling PRs (#1177/#1178/#1179); same file as #1162's fix but ~150 lines apart, disjoint hunks.

### DIFF
--- a/src/lib/merchantListStore.test.ts
+++ b/src/lib/merchantListStore.test.ts
@@ -490,6 +490,35 @@ describe("merchantListStore", () => {
 			expect(get(merchantList).totalCount).toBe(2);
 		});
 
+		// Consistency guard for the open/close boundary: the closed-panel badge
+		// and the open panel's list must report the same total for the same
+		// rows under every filter mode, so the count never jumps on open.
+		it("agrees with fetchAndReplaceList's totalCount for the same rows under every filter", async () => {
+			const recent = new Date(Date.now() - 30 * 86400000).toISOString();
+			const old = new Date(Date.now() - 2 * 365 * 86400000).toISOString();
+			const rows = [
+				{ id: 1, lat: 0, lon: 0, name: "A", verified_at: recent },
+				{ id: 2, lat: 0, lon: 0, name: "B", verified_at: old },
+				{ id: 3, lat: 0, lon: 0, name: "C" },
+			];
+
+			for (const filter of [null, 1, "outdated"] as const) {
+				merchantList.setVerifiedFilter(filter, { persist: false });
+
+				(api.get as Mock).mockResolvedValueOnce({ data: rows });
+				await merchantList.fetchCountOnly({ lat: 0, lon: 0 }, 10);
+				const closedPanelCount = get(merchantList).totalCount;
+
+				(api.get as Mock).mockResolvedValueOnce({ data: rows });
+				await merchantList.fetchAndReplaceList({ lat: 0, lon: 0 }, 10);
+				const openPanelCount = get(merchantList).totalCount;
+
+				expect(closedPanelCount, `filter ${String(filter)}`).toBe(
+					openPanelCount,
+				);
+			}
+		});
+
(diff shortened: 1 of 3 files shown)


escapedcat:
Copilot's typing finding adopted, with a stricter fix than suggested (87e1af98): the count response is now typed `Pick<Place, "id" | "verified_at">[]`, and rather than casting at the filter call, `filterPlacesByRecency` became generic over anything carrying `verified_at` — partial rows filter without pretending to be full Places, and every full-Place caller keeps its exact types.

escapedcat:
qodo's "badge ignores recency readiness" suggestion — **declined, with the trade now documented in code** (29f86a46). The `verifiedDatesLoaded` gate exists because the **bulk $places feed** carries no dates until enrichment — filtering it ungated would misclassify everything. `fetchCountOnly` counts **radius-API rows**, which carry real `verified_at` when requested, so filtering them is always correct — which is exactly why `fetchAndReplaceList` (the open-panel path, same API) filters ungated by design. Gating the badge as suggested would (a) desync it from the open panel's count, recreating the open/close jump this PR fixes, and (b) not fix the thing it aims at: during the one-time enrichment fetch the *markers* briefly show unfiltered pins — that's inherent to bulk-feed-vs-API data freshness, affects the open list identically, and self-heals when enrichment lands. The comment on `fetchCountOnly` now states this trade explicitly so it reads as a decision rather than an oversight.

PR #1181, opened 2026-08-01
fix(map): keep search-mode category counts in sync with the verified filter #1162

### LINKED ISSUE #1162: bug(map): search-mode category counts go stale after changing the verified filter

escapedcat:
Fixes #1162. Two commits, two halves of the same divergence:

## 1. Store: recompute counts on filter change
In search mode `applyVerifiedFilter`'s forced list refresh early-returns, so `categoryCounts` froze at search time while the rendered list re-derived reactively — chip enabled/disabled state and the "N of M" line disagreed with the visible rows. `setVerifiedFilter` now recomputes the counts from the **same recency-filtered selection the panel renders** (search rows carry their own `verified_at`, so no readiness gate — matching the panel and the marker block's `inSearch` bypass), reusing `applyCategoryFilter` so a window that zeroes the selected chip snaps selection back to All exactly as nearby mode does.

Deliberately scoped (analysis findings): nearby mode is untouched — the forced refresh already recomputes its counts via `setMerchants`/`fetchAndReplaceList`; the one nearby hole (closed-panel count path) is #1161's separate fix. Guarded on non-empty results so the empty search prompt and the `?outdated` mount-time seed stay no-ops. Fresh searches with a filter already active were already correct.

## 2. Panel: status line describes the rendered rows
The "N of M" branch required a category chip, so a verified-filter-only reduction reported raw `searchResults.length` — "100 results" above a 90-row list. The condition now fires on any length mismatch. Known trade (documented in the commit): truncated **and** filtered searches read shown-of-returned rather than returned-of-server-total, since the server total for the filtered set is unknowable client-side.

## Verification
- Unit (TDD, red→green): 8 new `setVerifiedFilter` cases — recompute per window, restore on clear, outdated inversion, auto-reset of a zeroed chip, keep of a populated chip, nearby untouched, empty-search untouched, persist contract. 73 store tests green.
- Live: "coffee" search (100 results) + "Verified within 1 year" → status "**90 of 100 result(s)**" with exactly 90 rows rendered; chips re-derive. Pre-fix: "100 of 1,091" above the same 90 rows.
- `format:fix` / `check` / `lint` / full suite clean.

## Merge order
Same file as #1180 but disjoint hunks (~150 lines apart); tests placed in a separate describe to avoid append conflicts. Any order with the other four.

### DIFF
--- a/src/lib/merchantListStore.test.ts
+++ b/src/lib/merchantListStore.test.ts
@@ -1,6 +1,6 @@
 import { get } from "svelte/store";
 import type { Mock } from "vitest";
-import { beforeEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import { CATEGORIES, placeMatchesCategory } from "$lib/categoryMapping";
 import {
@@ -92,6 +92,14 @@ describe("merchantListStore", () => {
 			lastUpdated: null,
 			usesMetricSystem: null,
 		});
+	});
+
+	afterEach(() => {
+		// Defensive: tests that persist the verified filter clean up inline,
+		// but a failing assertion would skip that — never leak the key. (Note
+		// reset() restores module-init state and does NOT re-read storage, so
+		// today a leak is inert; this guards the day that changes.)
+		localStorage.removeItem(VERIFIED_FILTER_STORAGE_KEY);
 	});
 
 	describe("state toggles", () => {
--- a/src/lib/merchantListStore.test.ts
+++ b/src/lib/merchantListStore.test.ts
@@ -6,6 +6,7 @@ import {
 	MERCHANT_LIST_FETCH_CEILING,
 	MERCHANT_LIST_MAX_ITEMS,
 } from "$lib/constants";
+import { VERIFIED_FILTER_STORAGE_KEY } from "$lib/map/verifiedFilter";
 import type { Place } from "$lib/types";
 
 // Mock the centralized axios instance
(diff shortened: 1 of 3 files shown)


escapedcat:
Copilot suggested reusing `applyCategoryFilter` here — declined as stated, because that would reintroduce what the previous review pass removed: it builds a filtered list nothing in this path consumes, via `filterMerchantsByCategory`, a different predicate implementation than the panel renders with (`placeMatchesCategory`). The legitimate kernel of the feedback is the drift risk of a re-implemented rule, and that's now solved properly (92cd8b8c): the auto-reset condition is extracted into a shared `shouldResetCategory` helper used by both `applyCategoryFilter` and `setVerifiedFilter` — one rule, no dead compute.

escapedcat:
qodo feedback dispositions (dbcdc2a3):

1. **localStorage test pollution** — adopted as a defensive `afterEach`, with one correction for the record: the described failure mechanism doesn't exist — `reset()` restores `initialState`, which captured `getStoredVerifiedFilter()` once at module import and is never re-read, so a leaked key is inert in this suite today. The hook guards the day that changes, and costs one line.
2. **"Misleading predicate comment"** — the comment was *accurate* (the two functions are different implementations: group icon-list scan vs last-write-wins `ICON_TO_CATEGORY` lookup, extensionally equal only while #1182's equivalence guard holds — the claim that they're "the same predicate" is exactly the assumption that guard exists to police). But two review bots misreading the same sentence means the sentence was unclear: it now names both functions and the guard explicitly.

PR #1182, opened 2026-08-01
test(map): guard the equivalence of the two category predicates #1159

escapedcat:
Consistency guard requested on top of the #1158–#1162 fixes.

The codebase filters by category through two predicates: `filterMerchantsByCategory` (group icon-list membership — used by the store's `applyCategoryFilter` and the nearby marker branch) and `placeMatchesCategory` (last-write-wins `ICON_TO_CATEGORY` lookup — used by the panel's `filteredSearchResults` and, after #1178, the search-mode pins). They are only extensionally equal **while no icon appears in two category groups**. The day that breaks, list and pins silently diverge again — the exact #1159 bug class.

Three new tests make that failure loud:
- the disjointness invariant itself, with a pointed failure message naming the offending icon and both groups
- predicate equivalence over every mapped icon plus unknown/empty/missing shapes, for every category
- `countMerchantsByCategory` agreement with both predicates

Passes on main today (no behavior change anywhere — pure guard). Consolidating to one predicate remains #1168's job; until then this is the tripwire.

Related consistency tests ride on the fix PRs where the behavior lives: #1180 gains "count path ≡ list path" totals, #1181 gains "chip counts ≡ rendered selection" oracles.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for category-based merchant filtering.
  * Verified consistent results across category matching, filtering, and counting.
  * Added validation for configured, unknown, empty, and missing category icons.

### DIFF
--- a/src/lib/categoryMapping.test.ts
+++ b/src/lib/categoryMapping.test.ts
@@ -420,4 +420,65 @@ describe("categoryMapping", () => {
 			expect(getIconColorWithFallback("some_unknown_icon")).not.toBe("");
 		});
 	});
+
+	// The map pins (search branch), the panel's filteredSearchResults, and the
+	// store's applyCategoryFilter split between filterMerchantsByCategory
+	// (group icon-list membership) and placeMatchesCategory (last-write-wins
+	// icon→category lookup). They are only extensionally equal while no icon
+	// appears in two groups — the moment that breaks, pins and list silently
+	// diverge again (the #1159 bug class). These tests make that divergence
+	// loud instead of silent.
+	describe("category predicate equivalence", () => {
+		// Every mapped icon, plus the shapes callers actually see: unknown,
+		// empty, and missing icons.
+		const corpus: Place[] = [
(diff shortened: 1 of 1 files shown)


escapedcat:
Copilot's finding adopted with one deviation (c95547a3): a duplicate icon within a single group now fails with an accurate message ("listed twice in group X — remove the duplicate") instead of the nonsense "in both restaurants and restaurants". Deliberately still a failure rather than ignored — a same-group duplicate doesn't break predicate equivalence, but it is always a config mistake worth catching; only the message needed fixing.

escapedcat:
qodo's order-sensitivity finding — adopted (2d902144): both selections are sorted before comparison, so the assertion now literally tests "the same set" (multiset — duplicates still counted). For the record, an order divergence is impossible today (both sides are `Array.prototype.filter` over the identical corpus array), but the sorted form keeps a reordering reimplementation from failing a test whose claim is set equality.

PR #1186, opened 2026-08-02
fix(area): grade the area map stars from the real report data #1163

### LINKED ISSUE #1163: bug(area): area map header star rating is always 5 stars
## Problem
The star rating in the area map header is a constant: every area shows 5 stars (or 0 for an empty area).

## Evidence
`src/components/area/AreaMap.svelte:481-486` `updateAreaGrade` sets `upToDate = total`, so `upToDatePercent` is always `"100"` and `getGrade` (`src/lib/utils.ts:152-164`) always returns 5.

## Fix direction
Either feed it real data — `areaReports[0].tags.up_to_date_percent`, which `AreaStats.svelte:42-44` already reads — or delete the stars.

---

escapedcat:
Fixes #1163.

## What was happening
`updateAreaGrade` in `AreaMap.svelte` set `upToDate = filteredPlaces.length` — the numerator equalled the denominator, so `upToDatePercent` was always `100` (or `0` for an empty area) and `getGrade` returned **5 stars for every area**. The header's own tooltip says 5 stars means ≥95% up-to-date; South Africa sits at **60%** and still showed five.

## The fix
The header now grades from the **latest report's `up_to_date_percent`** — the exact value the Stats section already renders — passed from `AreaPage`'s existing `areaReports` (no new fetches). The grade derives reactively; `updateAreaGrade` and its dead locals are gone.

Honest-degradation choice: while reports are loading — or for an area with no reports at all — the stars keep their existing pulsing placeholder rather than fabricating a rating. Showing nothing truthful beats showing five wrong stars, which is how this bug shipped in the first place.

## Verification
- Live: `/country/za` renders **3 filled + 2 dimmed** stars (60% → grade 3 band), previously 5; header settles out of the pulse state once reports land.
- `getGrade`'s percentage boundaries had zero test coverage — now pinned edge-by-edge (95/94, 75/74, 50/49, 25/24) since the visible stars finally depend on them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Area map grades now reflect the latest available area report percentage.
  * Grade stars remain in a loading state until map data and a valid grade are available.
  * Improved accuracy of percentage-to-grade boundary handling across the full range.

* **Tests**
  * Added coverage for grade calculations from 0% through 100%.

### DIFF
--- a/src/components/area/AreaMap.svelte
+++ b/src/components/area/AreaMap.svelte
@@ -76,8 +76,13 @@ const closeDrawer = () => {
 };
 
 let grade: Grade | undefined;
+// Round before grading so the stars agree with AreaStats, which displays the
+// same value via toFixed(0) — a fractional 94.6 must not show "95%" in Stats
+// while the header grades it below the 95% five-star boundary.
 $: grade =
-	upToDatePercent === undefined ? undefined : getGrade(upToDatePercent);
+	upToDatePercent === undefined
+		? undefined
+		: getGrade(Math.round(upToDatePercent));
 
 let gradeTooltip: HTMLButtonElement;
 
--- a/src/components/area/AreaMap.svelte
+++ b/src/components/area/AreaMap.svelte
@@ -33,6 +33,10 @@ import { browser } from "$app/environment";
 export let name: string;
 export let geoJSON: GeoJSON;
 export let filteredPlaces: Place[];
+// Latest report's up_to_date_percent for this area (same source AreaStats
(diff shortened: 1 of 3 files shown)


escapedcat:
qodo's rounding finding — adopted (edb3f538): the grade now derives from `Math.round(upToDatePercent)`, matching the convention AreaStats uses for display (`toFixed(0)`), so a fractional 94.6 can't read "95%" in Stats while the header grades it below the five-star boundary.

PR #1188, opened 2026-08-02
fix(sync): heal corrupt cached rows and guard the area-type filters #1187

### LINKED ISSUE #1187: bug(sync): corrupt cached rows crash the communities/countries/area pages
## Symptom
On https://btcmap.org/communities/africa (and any communities/countries/area page), affected browsers throw:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'type')
    at Array.filter (<anonymous>)
    at $$.update
```

and the page **silently renders an empty directory** — no toast, no error UI. Reproduced on production; confirmed the crashing site is the `area.tags.type === "community"` filter in `communities/[section]/+page.svelte`.

## Root cause (forensics)
The affected browser's IndexedDB cache (`BTC Map` → `areas_v4`) contained 916 rows, one of which was **the literal string `"\n"`**. Reading `.tags` on a string yields `undefined`; `.type` throws.

How a newline gets into a typed cache:
1. If the v2 API ever returns a **non-JSON response** (maintenance page, proxy error), axios hands the crawl a *string*.
2. `for (const item of newItems)` over a string iterates **characters**, merging them into the data set.
3. The dedup logic collapses them — every character has `id === undefined`, so each replaces the previous — leaving exactly **one surviving character: the response's trailing newline**.
4. The corrupt row then survives **every incremental merge forever**: the cache hydrate runs rows through `filterDeleted` (`!area.deleted_at && area.tags?.type !== "trash"`), which a string passes cleanly, and `incrementalSync` re-persists it each cycle.

Current API data is clean (verified: all 999 rows have proper tags) — the corruption lives in long-lived browser caches, so it hits **long-time visitors** while fresh profiles (and CI, which always starts clean) never see it. `merchantListStore` has the exact `Array.isArray` defense for this API failure mode; the sync factory never got it. `getCommunitiesAtCoordinates` even carries the comment "Some area records can come back without a tags object; guard before access" — the phenomenon was observed before, and one call site got a guard while five didn't.

## Why we didn't notice
- The failure is an uncaught rejection inside a Svelte reactive — no toast, no boundary, no telemetry; users just see an empty page.
- It only reproduces on caches that caught a transient API glitch — developers and CI always run fresh.
- Persisted cache data was implicitly trusted as `T[]` with no runtime validation at the trust boundary.

## Fix (two layers, one PR)
1. **Self-heal in `createSyncFactory`**: `isSyncableRow` validation at both trust boundaries — API responses (plus an `Array.isArray` check so string responses throw instead of char-merging) and cache hydration (poisoned caches heal on next load, for all four entity caches: areas/reports/events/users).
2. **Defense-in-depth guards**: the five unguarded `area.tags.type` reads become `area.tags?.type` (communities/[section], countries/[section], communities/map, AreaPage ×2), matching the existing guard in `getCommunitiesAtCoordinates`.

Verified end-to-end by injecting the exact `"\n"` corruption into a local cache: un-fixed code crashes with 0 cards rendered; fixed code renders 60 cards and re-persists a clean 915-row cache.

## Future-proofing (separate follow-ups worth considering)
- Client-side error telemetry (even a lightweight `unhandledrejection` beacon): this crash lived in the field for an unknown time and surfaced only because a user pasted a console error into chat.
- An e2e that seeds a corrupted cache before boot would pin this class permanently.

---
_Found investigating a user-reported console error on /communities/africa (2026-08-02)._

escapedcat:
Fixes #1187 — the user-reported crash on https://btcmap.org/communities/africa.

## The forensics
Affected browsers carry **the literal string `"\n"` as a row** in their persisted `areas_v4` cache. Reading `.tags` on a string → `undefined` → `.type` throws inside the communities filter, killing the reactive: the directory renders a **silent empty page**, forever, on every visit.

Origin: if the v2 API ever returns a non-JSON response (maintenance page, proxy error), axios hands the crawl a *string* — `for (const item of newItems)` iterates its **characters**, and the dedup collapse (every char has `id === undefined`) leaves exactly one survivor: the trailing newline. Once persisted, it passes `filterDeleted` (`tags?.type !== "trash"`) on every hydrate and survives every incremental merge. Current API data is clean — the corruption lives in long-lived caches, which is why fresh profiles and CI never reproduced it.

## The fix (two layers)
1. **`createSyncFactory`** — validate at both trust boundaries: reject non-array API responses outright (the same defense `merchantListStore` already had), run `isSyncableRow` over both crawls (raw count still drives pagination; cursor + merge use only valid rows, with a break so a garbage page can't spin on a stuck cursor), and **sanitize cache hydration so poisoned caches self-heal** on the next sync — for all four entity caches (areas/reports/events/users).
2. **Render guards** — the five unguarded `area.tags.type` reads become `area.tags?.type` (communities/countries directories, communities map, AreaPage ×2), matching the guard `getCommunitiesAtCoordinates` has carried for this exact observed phenomenon.

## Why nothing caught it (and what now does)
- No error telemetry, no toast: an uncaught rejection in a reactive is invisible.
- CI always starts with a fresh cache — stale-cache corruption is structurally unreachable.
- The existing communities e2e asserted title/buttons/chart but **never that cards render** — an empty directory passed CI.

New coverage: `communities.spec.ts` now asserts the directory actually populates, and `communities-corrupt-cache.spec.ts` poisons the cached areas with the exact field corruption, reloads, and demands the page still render (fails on pre-fix code: zero cards). Plus unit tests pinning `isSyncableRow` against the whole corruption class.

## Verification
- Reproduced the production crash (btcmap.org, readable stack on local dev), located the corrupt row live in an affected cache (index 908 of 916, value `"\n"`).
- Injected the identical corruption into a local cache: un-fixed → crash, 0 cards; fixed → **0 errors, 60 cards, cache re-persisted clean (915 rows, 0 corrupt)**.
- `format:fix` / `check` / `lint` / full unit suite green.

### DIFF
--- a/src/components/area/AreaPage.svelte
+++ b/src/components/area/AreaPage.svelte
@@ -172,7 +172,7 @@ const initializeData = async () => {
 		if (type === "community") {
 			return (
 				area.id === data.id &&
-				area.tags.type === "community" &&
+				area.tags?.type === "community" &&
 				area.tags.geo_json &&
 				area.tags.name &&
 				area.tags["icon:square"] &&
@@ -181,7 +181,7 @@ const initializeData = async () => {
 		} else {
 			return (
 				area.id === data.id &&
-				area.tags.type === "country" &&
+				area.tags?.type === "country" &&
 				area.id.length === 2 &&
 				area.tags.geo_json &&
 				area.tags.name &&
--- a/src/lib/sync/createSyncFactory.test.ts
+++ b/src/lib/sync/createSyncFactory.test.ts
@@ -1,0 +1,45 @@
+import { describe, expect, it } from "vitest";
+
+import { isSyncableRow } from "./createSyncFactory";
+
+describe("isSyncableRow", () => {
+	it("accepts a well-formed entity row", () => {
+		expect(
+			isSyncableRow({ id: "de", updated_at: "2026-01-01T00:00:00Z" }),
+		).toBe(true);
+		expect(isSyncableRow({ id: 42, updated_at: "2026-01-01T00:00:00Z" })).toBe(
+			true,
+		);
+	});
+
+	// The corruption observed in the field: a non-JSON API response parsed as
+	// a string gets character-iterated by the crawl loops, and the dedup
+	// collapse leaves one stray character (a trailing newline) that then
+	// survives every incremental merge in the persisted cache.
+	it("rejects stray string fragments", () => {
+		expect(isSyncableRow("\n")).toBe(false);
+		expect(isSyncableRow("")).toBe(false);
+		expect(isSyncableRow("<html>")).toBe(false);
+	});
+
+	it("rejects null, undefined, and primitives", () => {
+		expect(isSyncableRow(null)).toBe(false);
+		expect(isSyncableRow(undefined)).toBe(false);
+		expect(isSyncableRow(0)).toBe(false);
+		expect(isSyncableRow(true)).toBe(false);
+	});
+
+	it("rejects objects without a usable id", () => {
+		expect(isSyncableRow({})).toBe(false);
+		expect(isSyncableRow({ id: null, updated_at: "2026-01-01" })).toBe(false);
+		expect(isSyncableRow({ id: undefined, updated_at: "2026-01-01" })).toBe(
+			false,
+		);
+	});
+
+	it("rejects rows whose updated_at cannot drive the sync cursor", () => {
+		expect(isSyncableRow({ id: 1 })).toBe(false);
+		expect(isSyncableRow({ id: 1, updated_at: null })).toBe(false);
+		expect(isSyncableRow({ id: 1, updated_at: 12345 })).toBe(false);
+	});
+});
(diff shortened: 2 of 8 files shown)


escapedcat:
All three review findings adopted (807e9659):

1. **Validator soundness** (qodo + Copilot): `isSyncableRow` now requires a string/number `id` (the merge maps key on it) and a *parseable* `updated_at` (the pagination cursor depends on it) — plus test cases for object/array ids and non-date strings.
2. **Silent truncation** (qodo + Copilot): an empty page remains the normal end of data, but a non-empty page validating to zero rows now **throws** — `initialSync` surfaces it through the error store, `incrementalSync` propagates to the caller's fallback onto the sanitized cache. No more silently-partial syncs.
3. **IDB transaction semantics in the e2e** (qodo): the cache-poisoning write resolves on `tx.oncomplete` (with `onabort`/`onerror` rejection) instead of request success, since navigation can abort an uncommitted transaction.

Issue #1176, opened 2026-08-01
refactor(area): split AreaPage into per-section modules


escapedcat:
**Recommendation: Worth exploring.**

## Problem
`AreaPage.svelte` is one shallow mega-module: 582 lines, ~40 `let` declarations, a 23-variable `contact:*` fan-out spread back into `<Socials>` as 22 named props (`:211-258`, `:472-495`), three parallel section maps carrying no information (`:84-108`), a 13-line manual reset block (`:294-307`) that exists only because the component instance outlives the area — and an unconditional `batchSync` of four world crawls (`:62-64`) regardless of which section needs what (#1165 is the first slice of this). `description` from SSR is dropped; `iconSquare` is never read.

## Sketch
Route renders `{#key area.slug}` + `<AreaHeader {area}/>` + one section component per route; each section owns its data need (`reportsSync()` only inside stats). `<Socials contacts={area.contacts}/>` replaces the 23-field fan-out. The `{#key}` deletes the reset block wholesale.

## Wins
- Locality: "what does /stats need" is answerable by opening one file
- Cold load: 57.9 MB → ~2 MB for merchants/activity/maintain (with #1165 landed and stats owning reports)
- Testability: AreaHeader/SectionNav become prop-in/markup-out — the first unit-testable components in the repo

Depends on the loadAreaBundle and placesInArea issues; #1165 is its independent first slice. Note: `taggersFetchGeneration` (`:132-166`) is correct and load-bearing — carry it over untouched.

---
_Found by a Claude Code architecture review (2026-08-01, map-view focus)._

escapedcat:
**Re-review verdict: needs-update (still worth-exploring; sequenced LAST).**

- **Stale since #1186: "reportsSync() only inside stats" would break the merchants section** — the AreaMap stars now consume `areaReports?.[0]?.tags.up_to_date_percent` (AreaPage:540). The stars would pulse forever under the sketched split. Either per-section reportsSync in merchants too, or `up_to_date_percent` rides #1174's AreaBundle (coordinate there). The "57.9 MB → ~2 MB for merchants" number needs the same correction.
[...]

Issue #1174, opened 2026-08-01
refactor(area): loadAreaBundle — section-aware area loader


escapedcat:
**Recommendation: Strong.**

## Problem
`src/lib/areaSectionLoad.ts` validates `params.section` (`:44`) and then ignores it: every section gets the full `/v3/areas/{slug}` fetch (which includes the complete `geo_json` polygon and, for communities, ready-made `box:*` — verified against the live API) **and** a sequential `?limit=10000&offset=0` place-issues fetch (9.7 KB small community, **148.6 KB for the US**) — shipped in the SSR payload even to `/merchants`. Then the loader returns only `{id, name, tickets, issues, description}`, discarding the polygon — and the client re-downloads **5.7 MB of world areas** (`AreaPage.svelte:171-192` via the v2 crawl) to recover a 5.6 KB object the server already had. `offset=0` with no pagination loop silently truncates >10k issues. `box:*` is typed `string` but the API returns numbers (`types.ts:76-79`), and countries have no `box:*` at all.

## Sketch
```ts
// src/lib/area/load.ts
loadAreaBundle(fetch, slug, section): Promise<AreaBundle>
// hides: v3-vs-v4 endpoints, the 404/410/deleted ladder, per-section fetch pruning,
// issue pagination, bbox derivation (box:* else computeBbox), AreaTags → typed contacts object
```

## Wins
- Locality: adding a section = one row in a section spec; the issues cap lives in one place
- Leverage: callers get a typed area instead of `area["contact:satlantis"]` × 23
- Testability: `areaSectionLoad.test.ts` already mocks fetch per-URL — "no place-issues request for merchants" becomes a one-line assertion

---
_Found by a Claude Code architecture review (2026-08-01, map-view focus)._

escapedcat:
**Re-review verdict: SOUND — the only issue of the nine that survives as filed.** Every load-bearing claim re-verified on current main. This is the sequencing ROOT of the area cluster: #1174 → #1175 → #1176.

Three amendments to fold in before implementation:
1. **Pick one bbox convention**: the sketch's "box:* else computeBbox" mixes incompatible conventions — `computeBbox` is not antimeridian-aware (Fiji → useless full-width bbox) while `geoBounds` (utils.ts) returns wrap bboxes with west>east. Standardize on the geoBounds/wrap convention, documented, shared with #1175.
[...]

PR #1189, opened 2026-08-02
fix(sync): unstick the crawl cursor on shared timestamps #1164

### LINKED ISSUE #1164: bug(sync): v2 crawl loops forever if a full page of rows shares one updated_at
## Problem
The v2 sync crawl advances its cursor with an **inclusive** `updated_since = last row's updated_at` and loops `while (responseCount === limit)`. If a full page of rows ever shares one timestamp, the crawl re-fetches the same page forever.

## Evidence
- `src/lib/sync/createSyncFactory.ts:176,199` (initial crawl) and `:232,244` (update crawl)
- Not hypothetical-adjacent: the reports table has a bulk backfill where 10,000 rows span 6.8s (`2025-07-07T08:08:14.974Z` → `08:08:21.792Z`). Today's margin is ~1,470 rows/s against a 10,000 limit; nothing guarantees that margin holds for the next backfill.

Also worth noting while in there: the initial-crawl merge dedupes with `allData.findIndex` inside the page loop (`:186`) — quadratic on the duplicate path over 104k event rows.

## Fix direction
Strictly-greater cursor plus an id-based tiebreak, or detect a non-advancing cursor and bail. A test for "full page shares one timestamp" is currently unwritable because the crawl is entangled with localforage/stores — see the sync re-cut notes in the review.

---

escapedcat:
Fixes #1164.

## The hazard
The v2 `updated_since` parameter is **inclusive** — the proof is in the codebase itself: both crawl loops carry dedup logic that only exists because each page's last row reappears at the start of the next page. That inclusivity means a *full* page whose rows all share one `updated_at` can never advance the cursor: the crawl refetches the identical page until the tab dies. Not hypothetical-adjacent: the reports table already contains a backfill that wrote 10,000 rows in under 7 seconds — one bulk migration with identical timestamps away from every long-session user spinning on `/v2/reports` forever.

## The fix
`nextCursor(previous, pageLast)` — when the page's last timestamp equals the current cursor (only possible on the stuck-full-page case; a non-full page exits the loop regardless), nudge **one millisecond** past it. Rows sharing that timestamp beyond the page boundary are skipped for now and picked up whenever they next update: bounded, self-correcting staleness instead of an unbreakable loop. The alternative (bailing with an error) would stall the same user permanently — every retry hits the same wall.

Both crawl loops use the shared helper (per this PR set's no-drift convention), and #1188's `isSyncableRow` already guarantees the timestamp parses before it reaches the nudge.

## Also in the block
The initial crawl's dedup used `allData.findIndex` inside the page loop — quadratic on the duplicate path over 100k+ event rows (flagged in the same issue). It now uses the same insertion-ordered `Map` the incremental path already had; `Map.set` updates in place, so replace-at-original-position semantics are identical at O(1) per row.

## Verification
- Unit (TDD, red→green): cursor advances normally, nudges +1ms when stuck, rolls across second boundaries. 13 sync-factory tests green.
- A true end-to-end trigger needs the API to serve a full page of identical timestamps — not safely reproducible against prod; the helper *is* the decision point, and it's pinned directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronization pagination when multiple records share the same timestamp.
  * Prevented repeated fetches and sync loops at page boundaries.
  * Preserved correct handling of duplicate and deleted records during initial synchronization.
  * Added reliable cursor advancement across timestamp and second-boundary transitions.

### DIFF
--- a/src/lib/sync/createSyncFactory.test.ts
+++ b/src/lib/sync/createSyncFactory.test.ts
@@ -1,6 +1,10 @@
 import { describe, expect, it } from "vitest";
 
-import { isSyncableRow, validateSyncPage } from "./createSyncFactory";
+import {
+	isSyncableRow,
+	nextCursor,
+	validateSyncPage,
+} from "./createSyncFactory";
 
 describe("isSyncableRow", () => {
 	it("accepts a well-formed entity row", () => {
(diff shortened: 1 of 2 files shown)


escapedcat:
Both findings addressed in fd38319a — same core point from CodeRabbit and qodo, split disposition:

**Adopted (qodo's minimum ask): the deferral is no longer silent.** `nextCursor` now `console.warn`s with the endpoint and the stuck timestamp when a nudge fires, and the warn is pinned by test (including that normal advances stay silent).

**Declined: the compound `(updated_at, id)` cursor / page token.** Not a heavy lift — an impossibility against this API: `updated_since` + `limit` is the *entire* v2 pagination surface; there is no page token, offset, or id-tiebreak parameter to send. And the fallback idea fails too: a full re-crawl pages through the same API and hits the same unpaginatable tie at the same place. The nudge is the only progress-guaranteeing option this API admits, which the helper's comment now states explicitly. The deferral is bounded and self-correcting (rows return whenever they next update), and the longer-term exit is the planned migration off the v2 crawls entirely (#1174-#1176 direction).

Re qodo's reconciliation suggestion: a periodic full resync wouldn't recover the deferred rows either (same tie, same API) — the warn plus the eventual-update recovery is the honest ceiling here.

Issue #1208, opened 2026-08-04
refactor(svelte5): incremental runes migration — per-folder tracking


escapedcat:
Part of #1201 (Stage 4 — ongoing, **no deadline**: Svelte 5 legacy mode supports the old syntax indefinitely; runes/legacy components mix freely).

After Stage 1, the table cluster (11 files) is runes-mode and the other ~157 components still use Svelte-4 syntax. This issue tracks converting them folder by folder. `npx sv migrate svelte-5` supports per-folder runs and automates `export let`→`$props`, `$:`→`$derived`/`$effect` (wrapping timing-unsafe cases in `run()` from `svelte/legacy`), `on:`→event attributes, slots→snippets, `new Component`→`mount`.

Inventory (2026-08-04): `export let` in 130 files (406 uses), `$:` in 93 files (378), `<slot>` in 11 files, `<svelte:component>` in 4, `$$props`/`$$restProps` in 5, `createEventDispatcher` in 3.

## Approach

- One PR per folder below; run the migrator, then hand-review every `run()` wrapper and `@migration-task` comment it leaves.
- **Manual work the tool skips:** the 3 `createEventDispatcher` components (`AddressSearch`, `SaveAuthPrompt`, `ActivityTypeFilter`) → callback props, plus their ~6 consumers.
- **Hand-convert, don't auto-migrate, the reactivity hotspots:** `CommunityCard.svelte` (25 `$:`), `MerchantListPanel.svelte` (23), `merchant/[id]/+page.svelte` (20), `Socials.svelte` (19 `$:` + 23 props), `area/MerchantCard.svelte` (16). `routes/map/+page.svelte` is 2,189 lines but only 10 `$:` — size, not reactivity, is its risk; do it last and consider splitting first.

## Folder checklist

- [ ] `src/components/` (shared, excl. leaderboard — already runes from Stage 1)
- [ ] `src/components/area/`
- [ ] `src/routes/` — simple/static routes first
- [ ] `src/routes/merchant/`
- [ ] `src/routes/communities/` + `src/routes/countries/`
- [ ] `src/routes/map/` (last; 2,189-line page)
- [ ] sweep: remaining `<svelte:component>`, slots→snippets, `$$restProps` wrappers (`Icon`, `IconButton`, `CollapseButton`, `PrimaryButton`, `TextLink`)
- [ ] finale: enable `compilerOptions.runes: true` in `svelte.config.js` so legacy syntax can't creep back

[...]

PR #1271, opened 2026-08-22
fix(map): restore payment method URL filters #1269

### LINKED ISSUE #1269: Payment type filters on /map are not working
Payment type filters as described [here](https://gitea.btcmap.org/teambtcmap/btcmap-general/wiki/Embedding#if-you-would-like-to-filter-by-payment-method) are not working.
Some websites, like https://www.boltcard.org/ use it to embed filtered maps.

Options should be:

onchain
lightning
nfc

escapedcat:
Fixes #1269

## What broke

The `?onchain` / `?lightning` / `?nfc` params documented on the [Embedding wiki page](https://gitea.btcmap.org/teambtcmap/btcmap-general/wiki/Embedding#if-you-would-like-to-filter-by-payment-method) (and used by embeds like boltcard.org) were silently dropped in the #398 map rewrite (`be8955ad`) — the same regression class as `?boosts` (restored in #1097) and `?outdated` (restored in #1155). Nothing pinned their behavior, so nothing failed.

## What this does

- **`$lib/map/paymentFilter.ts`** — parse/match/serialize for the deep link, with the legacy Leaflet semantics: presence-only params, AND across methods, `nfc` → `payment:lightning_contactless`, and **strict `=== "yes"`** matching. The strictness is load-bearing: live OSM data carries `"no"` (13k places for onchain), `"Yes"`, `"yed"`, even email addresses in these tags.
- **`selectVisiblePlaces`** gains `paymentFilter`/`paymentReady` inputs, applied pre-category so chip counts describe the narrowed embed world. Search mode is exempt, same policy as boosts. The render signature covers both new inputs.
- **Lazy enrichment** — the bulk feed carries no payment tags, so a payment deep link fetches the three tags once via `ensurePaymentMethods()` (mirrors `ensureVerifiedDates`, including the degenerate-response latch guards; the filter stays inert until the tags land rather than hiding every pin). The tags also ride along on incremental `MAP_SYNC` rows so a mid-session update can't strip an enriched place out of the filter.
- **Enrichment lock** — combo deep links (`?issues&lightning`) dispatch both enrichers in one reactive flush; each does read→merge→publish over the whole `$places` array, so the later `places.set` could erase the earlier merge with its readiness flag already latched (blank embed for the session). Their merge windows now run serialized; fetches stay concurrent.
- **Map page wiring** — the deep link forces the local-markers list path (the radius API can't filter by payment) and pre-narrows the nearby list, mirroring boosts/issues.

## Tests (so this can't be silently lost again)

- Unit: parse/match/serialize semantics (incl. the `"no"`/noise rejection), pipeline narrowing + inert-until-ready gating + signature coverage, and the enrichment producer (latch guards, merge-by-id, no-op-once-latched, and a concurrency test that fails if the lock is removed — mutation-verified).
- e2e (`tests/map-payment-filter.spec.ts`, modeled on the `?outdated` guard): `?lightning` narrows the rendered pins; `?nfc` + AND semantics; and the nearby list narrows with the pins (San Salvador z15 anchor — ~250 places, ~12% nfc, wide drift headroom on both sides; new `__nearbyListCount` hook next to the existing `__mapPlacesCount` precedent).
- Thresholds calibrated from prod data: lightning=yes ≈79%, nfc=yes ≈12%, onchain=yes ≈40% of 29,320 places.

Not covered (noted deliberately): the search-mode exemption is comment-documented only, matching the equally-unpinned boosts exemption it copies.

## Notes for review

- `/communities/map` never had these params (verified in git history) — out of scope here; the wiki's "in addition to the location params above" phrasing slightly oversells, but the issue and boltcard.org are about `/map`.
- `routes/map/+page.svelte` and `MerchantListPanel` stay legacy syntax on purpose (#1208 hand-conversion hotspots).

## Summary by CodeRabbit

* **New Features**
  * Added payment-method filtering for embedded maps using `?onchain`, `?lightning`, and `?nfc` URL parameters.
  * Applied filters consistently to map pins and nearby-place lists, including combined filters.
  * Added lazy loading of payment information to support filtering without affecting search results.
* **Bug Fixes**
  * Improved enrichment reliability when payment and verification data load concurrently.
* **Tests**
  * Added coverage for payment filtering, deep links, combined criteria, readiness, and nearby-list behavior.

### DIFF
--- a/src/lib/api-fields.ts
+++ b/src/lib/api-fields.ts
@@ -49,14 +49,30 @@ export const PLACE_FIELDS = {
 		"osm:check_date:currency:XBT",
 		"osm:note",
 	] as const satisfies (keyof Place)[],
+	// The three tags the embed payment filter (?onchain&lightning&nfc) and
+	// the list's payment icons match on.
+	PAYMENT_METHOD: [
+		"osm:payment:onchain",
+		"osm:payment:lightning",
+		"osm:payment:lightning_contactless",
+	] as const satisfies (keyof Place)[],
 } as const;
 
 export const PLACE_FIELD_SETS = {
 	// verified_at rides along on incremental updates so a changed place keeps
 	// its date through a sync (no flicker) and the recency filter never needs a
 	// full re-fetch to stay fresh — the baseline is loaded once, lazily, by
-	// ensureVerifiedDates() when the filter is first engaged.
-	MAP_SYNC: [...PLACE_FIELDS.CORE, ...PLACE_FIELDS.SYNC, "verified_at"],
+	// ensureVerifiedDates() when the filter is first engaged. The payment tags
+	// ride along for the same reason: incremental rows replace enriched rows
+	// wholesale, and without them a mid-session update would silently drop a
+	// place out of a payment-filtered embed (ensurePaymentMethods() loads the
+	// baseline lazily, like the dates).
+	MAP_SYNC: [
+		...PLACE_FIELDS.CORE,
+		...PLACE_FIELDS.SYNC,
+		"verified_at",
+		...PLACE_FIELDS.PAYMENT_METHOD,
+	],
 	COMPLETE_PLACE: [
 		...PLACE_FIELDS.CORE,
 		...PLACE_FIELDS.SYNC,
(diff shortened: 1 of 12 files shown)


escapedcat:
Superseded — #1269 was fixed by #1272 plus the stacked hardening in #1274 (both merged), which together carry everything from this PR. via Claude Code 🤖

PR #1292, opened 2026-08-27
feat(map): duplicate-place interrupt before add-location handoff

escapedcat:
## Motivation

image

With the `submit_place` pipeline live (#1279), every duplicate submission now costs a supertagger a Gitea ticket round-trip to catch and close. This is step 3 of the add-location redesign ladder, following #1276, #1277, #1279, and #1289: before a placed pin hands off to the add-location form, check whether it's likely a duplicate and give the user a chance to link to an existing place instead of filing a new one.

## Behavior

- On placement confirm, existing (non-deleted) places within a 75 m radius of the pin are surfaced, closest first, capped at 5 candidates.
- If none are found, confirm goes straight to the form as before.
- If candidates exist, an interrupt sheet ("Is it one of these?") lists them; each candidate links to its merchant page so the user can check it in a new context without losing placement state.
- **Back** returns to the placement pin sheet.
- **None of these — add a new place** ("add-anyway") continues to the form with the pin's coordinates.
- Moving the map while the interrupt is open resets the candidate list (it's re-derived from the pin's current position).
- Duplicate detection needs no new API calls: candidates are computed client-side from the map's existing places store, which deliberately carries no `name` field (map-perf constraint). Candidate names are the one exception — they're fetched at interrupt-show time via the existing `/v4/places/search/` radius endpoint, falling back to "Unnamed place" if the lookup fails or returns nothing.
- If the places store is empty or still syncing, no candidates are found and confirm goes straight to the form — the same behavior as before this feature.

## Analytics

- `add_place_nearby_shown` — fires when the interrupt is shown, with a `count` of candidates.
- `add_place_nearby_continue` — fires when the user picks add-anyway.
- `add_place_nearby_candidate_click` — fires when the user opens a listed candidate's merchant page; this is the feature's success outcome, a duplicate averted.
- `add_place_confirm` keeps its existing meaning: "handed off to the form." It now fires on both the direct path (no candidates) and the add-anyway path, so the funnel metric is unchanged by the interrupt.

## i18n

The whole `map.placement` namespace (step-1 confirm sheet + this interrupt) is now translated in all 8 non-English locales, matching each file's tone and diacritic conventions — translations are machine-assisted, native-speaker review welcome.

## Testing

New spec `tests/map-placement-dedupe.spec.ts` covers all three paths (interrupt shown + Back, add-anyway, no-nearby-places) against the hermetic `stubMapData` fixture. Full checklist (format, check, typecheck, lint, unit tests) and the full Playwright suite are green.

https://claude.ai/code/session_01NjZ7qPmjDm85vTn9MbS2uS

## Summary by CodeRabbit

* **New Features**
  * Added nearby-place checks when placing a pin to help prevent duplicate locations.
  * Nearby matches now show distance information, with options to open an existing place, go back, or continue adding a new one.
  * Added merchant address copying with confirmation feedback.
  * Added placement-flow support across English, Bulgarian, German, Spanish, French, Italian, Dutch, Brazilian Portuguese, and Russian.

* **Tests**
  * Added coverage for proximity filtering, result limits, deleted places, empty results, name lookup failures, and placement-flow navigation.

### DIFF
--- a/src/lib/analytics.ts
+++ b/src/lib/analytics.ts
@@ -46,7 +46,9 @@ export type EventName =
 	| "backup_modal_shown"
 	| "backup_credentials_copied"
 	| "add_place_enter"
-	| "add_place_confirm";
+	| "add_place_confirm"
+	| "add_place_nearby_shown"
+	| "add_place_nearby_continue";
 
 declare global {
 	interface Window {
--- a/src/lib/analytics.ts
+++ b/src/lib/analytics.ts
@@ -48,7 +48,8 @@ export type EventName =
 	| "add_place_enter"
 	| "add_place_confirm"
 	| "add_place_nearby_shown"
-	| "add_place_nearby_continue";
+	| "add_place_nearby_continue"
+	| "add_place_nearby_candidate_click";
 
 declare global {
 	interface Window {
--- a/src/lib/i18n/locales/bg.json
+++ b/src/lib/i18n/locales/bg.json
@@ -725,15 +725,15 @@
 	"map": {
 		"bitcoinMerchantMapTitle": "Карта на биткойн търговци",
 		"placement": {
-			"title": "Place the pin",
-			"hint": "Drag the map until the pin sits on the business entrance, then confirm.",
-			"cancel": "Cancel",
-			"confirm": "Add a place here",
-			"nearbyTitle": "Is it one of these?",
-			"nearbyHint": "These places already exist near your pin — open one to check, or add a new place.",
-			"nearbyUnnamed": "Unnamed place",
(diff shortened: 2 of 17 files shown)


bubelov (approved):
I'd significantly increase the radius, otherwise concept ack. When I was an active editor I often found duplicates up to 1 km away, people confuse streets all the time, especially in boring cities where it all looks like a monotonic grid

escapedcat:
> I'd significantly increase the radius, otherwise concept ack. When I was an active editor I often found duplicates up to 1 km away, people confuse streets all the time, especially in boring cities where it all looks like a monotonic grid

👍 see stacked PR

PR #1297, opened 2026-08-29
feat(merchant-drawer): add copy button to the merchant address

escapedcat:
## Why

People want the merchant address as text so they can paste it into their own maps or directions app.

Today the address is a plain `<p>`. On mobile that text is effectively un-selectable: it sits inside a bottom sheet whose `pointerdown` / `touchmove` handlers fight long-press-to-select, so there is no reliable way to get the address out of the drawer.

## What

A small clipboard button after the address, swapping to a check for ~2s on success.

Because it lands in `MerchantDetailsContent.svelte`, it shows up in all three places that component backs:

- the desktop map drawer
- the mobile map drawer
- the area merchant drawer

```
 ┌────┐  Bitcoin Cafe
 │ ☕ │  Hauptstr. 12, 20095 Hamburg  ⧉
 └────┘  ● Open now
```

## Commits

1. **`refactor(copy-button)`** — `CopyButton` was hardcoded to a 24px FontAwesome clipboard in link colour with **no accessible name**. Parameterised `size`, `class` and the `aria-label`/`title`, converted to runes mode, and cleared the reset timeout on destroy. Defaults keep the existing invoice and donation call sites rendering identically.
2. **`feat(merchant-drawer)`** — the address affordance itself, plus `merchant.copyAddress` / `merchant.addressCopied` added to all 9 locales.
3. **`refactor(copy-button)`** — settle on one copy icon app-wide (see below).

## Details

- **One copy icon everywhere: material `content_copy`.** `BackupCredentials` already uses it twice, and material is the dominant set in the repo (91 uses vs 55, most of the FA ones being brand glyphs). `CopyButton`'s FA `clipboard` was the only one of its kind, so the two existing call sites move to `content_copy` too. One visible side effect: in `DonationOption.svelte:52` the copy button now sits next to an FA `qrcode`, so an outlined icon beside a filled one — minor, and worth it for a single copy affordance app-wide.
- 24px tap target (`h-6 w-6`) around the 14px mark, negatively offset so it stays optically aligned with the first line of a wrapping address.
- Muted `text-body/60`, so it reads as a secondary affordance rather than competing with the merchant name link.
- The address text stays plain selectable text — the button is additive.

## Testing

`format:fix` · `check` (0 errors) · `typecheck` · `lint` · `test --run` (692 passing) all green. No component test harness exists in this repo yet (vitest only covers `src/lib`), so the interaction wants a look on the deploy preview.

## Notes / follow-ups

- **`shareMerchant` can show a false success.** `src/lib/utils.ts:235` fires `navigator.clipboard.writeText` with no `await`, no `.catch()` and no fallback, while both callers (`MerchantDetailsContent.svelte:267`, `MerchantActionChips.svelte:20`) show their ✓ confirmation unconditionally. If the write rejects, the user sees "copied" with an empty clipboard, plus an unhandled rejection. Not touched here — separate PR.
- **Three clipboard implementations at three reliability tiers** — `CopyButton` (API + `execCommand` fallback + success gate), `BackupCredentials:17` (API + toast, no fallback), `utils.ts:235` (fire-and-forget). Worth extracting to `$lib/clipboard.ts`; as a bonus that puts the logic under vitest's `src/lib` coverage.
- **`MerchantDetailsContent.svelte` left in legacy mode.** The boy-scout rule in `CLAUDE.md` points at converting it, but this is a ~12-line addition to a 388-line file backing three drawers — an unverified whole-file conversion felt like the wrong thing to bundle with it. Happy to do it separately.
- **The `Navigate` action is `geo:{lat},{lon}`** (`MerchantDetailsContent.svelte:229`). Works on Android, does nothing on iOS Safari or desktop — plausibly *why* people are reaching for copy/paste. A universal `https://` maps URL with name + address would cover every platform.
- The two pre-existing `CopyButton` call sites (invoice, donation) still pass no `label`, so those buttons remain unnamed for screen readers. Each would need its own i18n key.

### DIFF
--- a/src/components/CopyButton.svelte
+++ b/src/components/CopyButton.svelte
@@ -13,7 +13,6 @@ type Props = {
 	label?: string;
 	// Announced/shown in place of `label` while the copy confirmation is up.
 	copiedLabel?: string;
-	iconType?: "fa" | "material";
 	class?: string;
 };
 
@@ -22,11 +21,11 @@ let {
 	size = "24",
 	label = undefined,
 	copiedLabel = undefined,
-	iconType = "fa",
 	class: className = "text-link transition-colors hover:text-hover",
 }: Props = $props();
 
 let copied = $state(false);
+const currentLabel = $derived(copied ? (copiedLabel ?? label) : label);
 let resetTimeout: ReturnType<typeof setTimeout> | undefined;
 
 const copy = async (field: string) => {
@@ -59,13 +58,13 @@ onDestroy(() => clearTimeout(resetTimeout));
 <button
 	type="button"
 	class={className}
-	aria-label={copied ? (copiedLabel ?? label) : label}
-	title={copied ? (copiedLabel ?? label) : label}
+	aria-label={currentLabel}
+	title={currentLabel}
 	onclick={() => copy(value)}
 >
 	{#if copied}
-		<Icon type={iconType} icon="check" w={size} h={size} />
+		<Icon icon="check" w={size} h={size} />
 	{:else}
-		<Icon type={iconType} icon={iconType === 'material' ? 'content_copy' : 'clipboard'} w={size} h={size} />
+		<Icon icon="content_copy" w={size} h={size} />
 	{/if}
 </button>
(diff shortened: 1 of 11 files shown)


dadofsambonzuki:
Great addition. This was bugging me the other day!

dadofsambonzuki (approved):

Issue #1134, opened 2026-07-15
feat: guided add-location wizard (physical / online-or-mobile, OSM detection, approve-to-push)


jurjendevries:
## Summary

Add a guided, multi-step **add-location wizard** (new route, alongside the existing `/add-location`) that adapts the questions to the kind of business and makes it easy for non-technical submitters to contribute good data.

## Motivation

The current `/add-location` form is a single long page that assumes a physical, OSM-style merchant. It doesn't fit:

- **Online / mobile businesses** — webshops and services delivered at the customer's place (e.g. a mobile window cleaner). These have no visitable location, so a map pin + address makes no sense.
- **Businesses already on the map** — no easy path to *update* an existing merchant's details.
- **Non-Western addressing** — many regions describe a location rather than using a street address.

## Proposed flow

```
Q: Does the business have a physical location visitors can go to?  [Yes] [No]

No  ──► Online-or-mobile form (name, website, socials, category, payment, contact)
        → stored for a future online/mobile listing (label: online-or-mobile)

Yes ──► "Is your business already on the map?"  (OSM map + search)
         ├─ select an existing merchant  → update form (website, phone, hours, socials, payment)
         ├─ search result "This is my business" → new form, prefilled from OSM POI data
         └─ "It's not on the map yet"    → new form (blank)
```

Key UX points:

- **Address _or_ location description** — one required, order adapts to locale (some regions describe rather than address).
- **Friendly opening-hours editor** that outputs OSM `opening_hours` syntax, incl. multiple ranges/day (lunch breaks) and 24/7.
- Search results prefill name / category / website / phone / opening hours from OSM (Nominatim `extratags`/`namedetails`).
- Browser back/forward navigates between steps (shallow routing).

## Data flow & spam control

- Every submission creates a **Gitea issue** (existing captcha + honeypot spam gate) with a machine-readable payload block.
- On maintainer approval (label), a webhook can push the data:
  - **btcmap-api `submit_place`** → instant map draft (idempotent by `gitea-<issue>`), and
  - **OSM changeset** via a dedicated bot account.
- Everything is env-gated: without tokens the wizard still works and just creates Gitea issues.

## Notes / open questions

- New Gitea labels needed: `online-or-mobile`, `update-location`, `osm-approved`.
- Requires ops setup for the automated push: a btcmap-api `places_source` token + import origin, and (new to the org) an OSM bot account + OAuth2 token. There is currently no automated OSM writer anywhere in the org.
[...]

PR #1299, opened 2026-08-31
feat(add-location): redirect to placement mode without pin coords #1134

escapedcat:
## Summary

Try:
image

https://deploy-preview-1299--btcmap.netlify.app/map?add=nav

image
image

Step 4 of the add-location redesign (#1134): **one placement UI in the system.** `/add-location` without valid `?lat&long` now redirects to the map's placement mode, so every add-place path — header, home, area pages, PWA shortcut, old bookmarks — goes through pin placement and the duplicate-check interrupt from #1292. Follows the radius review on that PR, which was really about the paths that bypassed the check.

- **Load guard** (`src/routes/add-location/+page.ts`): universal `load`, missing/invalid coords → `302 /map?add=redirect`. Works for SSR and client-side navigation; old links redirect rather than 404.
- **The page collapses to its details-first state**: arrival banner always on, marketing hero gone, and the location section is a non-interactive minimap preview of the confirmed pin (`StaticMapPreview`, promoted from the merchant hero's `MerchantStaticMap`) with the placement pin glyph overlaid. The whole box links back into placement mode centred on that pin (`/map?add=adjust#17/lat/long`). `StaticMapPreview` gained a `zoom` prop for this (the merchant hero keeps its 15). Inputs gate on the captcha only; "Submit another" returns to placement mode. Address search, the click-to-place picker, marker plumbing and the advanced coordinate inputs are deleted (960 → ~630 lines). `AddressSearch` and `$lib/geocoding` had no other users and are removed.
- **Entry points** (header dropdown, home CTA, area page, map menu fallback row, PWA manifest shortcut — with its "locaion" typo fixed in passing — and `llms.txt`) point at placement mode directly. In code the URL is spelled once (`placementEntryUrl(entry)`, typed by the `AddEntryMethod` union: nav / redirect / adjust / shortcut / another, anything else → `url`); the `/map` route's `+page.ts` load captures the value per navigation — before `AddPlaceMode`'s URL effect normalises shared URLs to a bare `?add=` — and reports it as the `add_place_enter` method so the funnel can tell entry points apart. `llms.txt` links a bare `?add` so agent traffic counts as `url`.
- **i18n**: the 14 orphaned `addLocation.*` keys and `forms.selectLocation` are removed from all nine locales, plus the `addressSearch` namespace from the three that had it (en/it/nl); key parity with `en.json` preserved.
- Service worker no longer caches redirected responses (`Cache.put` rejects them, and `/add-location → /map` would otherwise be cached under the wrong URL).

## Decisions worth a second look

- **No non-WebGL fallback.** Previously a browser without WebGL could still submit via the advanced coordinate inputs; now the only way in is the map, which itself shows the unsupported notice. Agreed as a product decision for this step (placement needs the map anyway) — flagging it here so it's explicit rather than absorbed into a refactor.
- **Runes conversion of `add-location/+page.svelte` is deliberately a follow-up PR** — a whole-file syntax conversion on top of a −800-line feature diff would be unreviewable. The file is now small enough to convert cleanly.

## Tests

- Unit: load guard (valid passthrough; missing / half / out-of-range / non-numeric → redirect), `buildPlacementUrl`, `placementEntryUrl`, `addEntryMethod`, the `/map` load (entry capture + server data forwarded).
- E2E: `add-location-coords.spec.ts` deleted with the feature; `add-location-arrival.spec.ts` rewritten (banner, minimap link href, click-through into placement mode on the pin, plain and malformed `/add-location` land in placement mode — hermetic via `stubMapData`); slim-form gotos carry coords; `map-menu` asserts the reload link; `home` CTA lands in placement mode.
- `pnpm run check` 0 errors · `typecheck` clean · `lint` clean · 710 unit tests · affected e2e specs green locally (arrival, slim-form, dedupe, map-menu, home, map-drawer, merchant-list-panel, merchant page).

## Follow-ups (not in this PR)

- Runes conversion of the add-location page (own PR).
- Interrupt sheet a11y (`role="dialog"`, live-region announcement, focus restore on map-drag dismissal).
- Then the roadmap continues towards hosting the form in a map side-panel — this PR is step 1 of that (the form can no longer carry a second map).

## New Features
- Added map-based location placement from the homepage, navigation menus, and PWA shortcuts.
- Location previews now support configurable zoom levels.
- Placement tracking now distinguishes entry points such as navigation, redirects, adjustments, and shortcuts.

## Bug Fixes
- Invalid or incomplete location links redirect to map placement.
- Redirect responses are no longer cached for offline use.
- Corrected the PWA shortcut description.

## Updates
- Retired address search and manual coordinate entry from the add-location flow, including related translations.


dadofsambonzuki:
1) Button looks a little excessive!

image

2) Local name should be more explicit, like "Name of place in local language".

image

3) Address is hidden behind extra details. Given the current discussion in the [Supertaggers Room](https://matrix.to/#/#btcmap-taggers:matrix.org) we should probably pre-populate this by reverse geocoding from the location and asking for a confirmation. [OSM Nominatim](https://nominatim.openstreetmap.org/ui/reverse.html) could be useful.

escapedcat:
> 1. Button looks a little excessive!

Yeah, it's just POC for mible. Adjusted a bit in the stacked PR. Will be a modal in the near future anwyays.

>     2. Local name should be more explicit, like "Name of place in local language".

See stacked PR

>     3. Address is hidden behind extra details. Given the current discussion in the [Supertaggers Room](https://matrix.to/#/#btcmap-taggers:matrix.org) we should probably pre-populate this by reverse geocoding from the location and asking for a confirmation. [OSM Nominatim](https://nominatim.openstreetmap.org/ui/reverse.html) could be useful.

Will do in a follow up via #1315

PR #1316, opened 2026-09-03
feat(add-location): pre-populate the address by reverse geocoding the pin #1315

### LINKED ISSUE #1315: Add-location: pre-populate the address by reverse geocoding the pin
Part of the add-location redesign (#1134), item 3 of its UX backlog. Raised in the review of #1299 and in the current Supertaggers discussion about spam ("make the address field mandatory").

## Problem

Since #1299 every `/add-location` arrival carries map-confirmed coordinates, but the address field is optional and hidden behind the "Add more details" expander — so most submissions arrive without an address, and typing one manually is friction nobody wants.

## Proposal

- On form arrival, fire **one Nominatim reverse-geocode** request for the pin coordinates and pre-fill the address input with the result.
- Move the address field **out of the expander**, with a short hint asking the submitter to confirm or correct it ("check this matches the entrance").
- The field stays editable; if the lookup fails it simply stays empty and optional — exactly today's behavior.
- Once pre-filled, marking the field `required` costs legitimate submitters nothing — which answers the mandatory-address ask with far less friction than forcing manual entry.

## Notes

- Nominatim usage policy is comfortably met: a single request per form arrival, identification via Referer, OSM attribution shown near the field.
- No API change needed for the pre-fill itself — the address already travels as `extra_fields.address` in `submit_place`. Enforcing address-mandatory server-side (for all origins, not just the website) would be a separate btcmap-api follow-up.

escapedcat:
Closes #1315.

image

The address field is promoted out of the "Add more details" expander and sits directly under the pin preview it derives from. On every fresh form arrival (the load guard guarantees pin coords, adjust round-trips remount with new ones) the page fires **one Nominatim reverse-geocode** and pre-fills the field.

- **Compact suggestion**: `road number, postcode settlement` composed from the structured parts (most specific settlement wins), `display_name` only as fallback — not the full region/country chain.
- **Confirm, don't trust**: a helper line under the label says the address is suggested from OpenStreetMap and asks to check it against the business entrance — doubling as the attribution Nominatim's policy expects. While the lookup is pending the placeholder reads "Looking up the address…".
- **Conditional required** (as discussed): a fulfilled suggestion flips the field to `required` — it can be corrected but not blanked out, which answers the mandatory-address ask at zero cost to legit submitters. Any failure (HTTP error, timeout, unresolvable point) leaves the field empty and optional, exactly the previous behavior.
- **Service worker**: `nominatim.openstreetmap.org` joins the never-intercept list — the old catch-all would have cached lookup responses into the app cache and served stale addresses offline.
- **Analytics**: `add_place_address_prefill` with `outcome: hit|miss`, so the funnel shows how often OSM actually has the address.
- **`moreDetailsToggle` copy** stops advertising the address in all nine locales; `addressLabel`/`addressTooltip` are replaced by the new hint/pending keys (key parity kept).

Testing: 7 new unit tests for `reverseGeocode` (param shape, composition, settlement preference, fallbacks, error paths); new Playwright spec for the hit path (value, required, hint, still editable) and the miss path (empty, optional); the existing add-location specs stub the lookup so they stay hermetic. Full suite green locally.

Per the issue discussion: fires on first arrival from in-line map placement (not only adjust round-trips), and the field is promoted out of additional fields.

## Summary by CodeRabbit

- **New Features**
  - Add-location now suggests an address from the confirmed map pin and selected language.
  - Suggested addresses appear beneath the map, remain editable, and become required when found.
  - Lookup status, OpenStreetMap attribution, and localized guidance are displayed.

- **Improvements**
  - Manually entered addresses are preserved during lookup.
  - Address lookups use the network for current results.
  - Added analytics tracking for lookup outcomes.

- **Tests**
  - Added coverage for address suggestions, validation, fallback behavior, and lookup failures.


dadofsambonzuki:
Very nice.

dadofsambonzuki (approved):

PR #1333, opened 2026-09-04
feat(map): host the add-location form in a side panel on the map #1134

escapedcat:
## What

The in-map host for the add-location form ([#1134](https://github.com/teambtcmap/btcmap.org/issues/1134)) — confirming a placement pin now opens the form **on the map** instead of navigating to `/add-location`:

- **Desktop:** right-side panel (MerchantDrawer sizing, attribution corner left uncovered). The map and crosshair pin stay live — every settled move hands the new center to the form, whose address lookup re-runs (the coords-reactivity from #1332 doing its job).
- **Mobile:** full-screen sheet.
- **History/URL:** opening pushes an entry, so browser Back walks form → placement (the merchant-drawer idiom); `?add=form` makes the state shareable and reload-safe with the pin in the hash. Escape and the × close the same way. The load's entry-method parsing folds `form` into `url`, so the funnel dimension stays bounded, and `add_place_confirm` still fires at the handoff.
- **Success in-panel:** check screen with "Submit another" (back to placement, tracked as `add_place_enter: another`) and a new "Back to the map" exit (i18n key ×9 locales).
- `AddLocationForm` gains `showPinPreview` — the panel hides the static minimap (the live map is the preview); the standalone page keeps it.
- Rides along: the merchant list panel is now **hidden, not unmounted**, during placement — Cancel no longer loses search state.

## Screenshots

**Desktop — the merchant drawer's dialect: left-docked card, map controls stay visible**

Right-side panel over the map with the crosshair pin visible and the address field prefilled

**Mobile — full-screen sheet**

Full-screen add-location sheet on mobile

## What's next (PR 3)

`/add-location` with coords redirects into this host and the standalone page retires — after this proves itself.

## Testing

- New `map-add-form.spec.ts`: confirm opens the form in place, × returns to placement, browser Back walks back, reload restores `?add=form` at the hash pin
- `map-placement-dedupe.spec.ts` updated: add-anyway / no-nearby now assert the in-map form (URL stays `/map`)
- All 21 merchant-list-panel + placement specs green; svelte-check / tsc / biome clean; 749 unit tests
- Verified visually against the dev server (screenshots above)

## Summary by CodeRabbit

- **New Features**
  - Add locations directly within the map using a mobile-friendly form sheet or desktop side panel.
  - Keep the map, placement pin, and selected coordinates visible while completing the form.
  - Add options to add another location or return to the map after submission.
  - Support browser back/forward navigation and reopening the form from a shared map URL.
  - Add a configurable option to hide the form’s static pin preview.
  - Add localized “Back to the map” labels across supported languages.

- **Bug Fixes**
  - Preserve merchant list search state during placement.
  - Ensure Escape and Cancel return users to the placement flow correctly.


dadofsambonzuki (approved):

dadofsambonzuki:
As a follow up, we should add a login (BTC Map / Nostr) option instead of / alongside the public email contact.

That will help us prioritize submissions based on submission history and/or web-of-trust.

escapedcat:
> As a follow up, we should add a login (BTC Map / Nostr) option instead of / alongside the public email contact.
> 
> That will help us prioritize submissions based on submission history and/or web-of-trust.

https://github.com/teambtcmap/btcmap.org/pull/1339

PR #1351, opened 2026-09-08
feat(area): add events section to community and country pages

bubelov:
Adds an 'events' tab between merchants and stats on /community/[area] and /country/[area], listing the meetups and conferences inside the area's polygon (last 365 days + all future, future first). Each event links to its website, past events are greyed out and tagged '(past)', the events tab shows a future-event count badge.

Backend: GET /v4/areas/{id_or_alias}/events, fetched server-side only for the events section to avoid the second round-trip elsewhere. The window is 'now - 365d' to 'now + forever'; soft-deleted and out-of-window events are dropped server-side. The /v4 Item struct lacks a #[ts(export)] annotation yet, so AreaEvent lives in $lib/types.ts until ts-rs is wired up on the btcmap-api side.

Datetimes are displayed verbatim as written by the organizer: the regex parser drops any timezone suffix (Z or +/-HH:MM) and treats the remaining wall-clock value as local time, so '19:00' always prints as '19:00' regardless of the upstream offset or the viewer's tz.

## Screenshots

Deploy preview at c843666f.

/country/th/events on desktop: Events (3) tab badge, two dated upcoming meetups, a Date TBD entry, then greyed (past) events

/community/chiang-mai/events in dark mode: Events (1) badge, one upcoming meetup and one past multi-day event

/country/br/events empty state: Events (0) badge and the no-events message

/country/th/events on a 390px-wide mobile viewport

🤖 Generated with [opencode](https://opencode.ai)

## Summary by CodeRabbit

- **New Features**
  - Added Events pages for community and country areas.
  - Displays upcoming and past events with dates, times, website links, and past-event labels.
  - Shows an empty state when no events are available.
  - Events navigation now displays the number of upcoming events.
  - Added localized Events labels and messages across supported languages.

- **Bug Fixes**
  - Preserves displayed local event times and reliably distinguishes past events.
  - Sorts upcoming events before past events.
  - Handles unavailable, incomplete, or missing event data gracefully.


escapedcat:
Two things I'd fix before merging, plus a note on CI:

1. **The events fetch now 502s every area tab.** Since c843666f `fetchAreaEvents` runs for every section and throws 502 on any non-404 / non-array response:

https://github.com/teambtcmap/btcmap.org/blob/c843666f28aac5c70c61f61d9725901f162122db/src/lib/areaSectionLoad.ts#L121-L131

https://github.com/teambtcmap/btcmap.org/blob/c843666f28aac5c70c61f61d9725901f162122db/src/lib/areaSectionLoad.ts#L204-L207

So an events-endpoint hiccup takes down merchants/stats/activity/maintain for every community and country, where before only the events tab depended on it. Suggest degrading to `[]` outside the events section (badge shows 0) and only 502-ing on the events tab itself. The PR description still says events are fetched "only for the events section", worth updating too.

2. **`href={event.website}` is unsanitized.** It's scraped third-party data, so run it through `safeHttpUrl` (same as the favicon in `community/[area]/+layout.svelte`) and fall back to the plain-text branch when it isn't http(s):

https://github.com/teambtcmap/btcmap.org/blob/c843666f28aac5c70c61f61d9725901f162122db/src/components/area/AreaEventsSection.svelte#L136-L146

CI: the red e2e check is unrelated to this PR. Since 2026-09-07 `main` fails the same 23 map tests because `api.iconify.design` SVG responses reach the GitHub runners without CORS headers; I'm looking into that separately.

via Claude Code 🤖

escapedcat:
Fair points, and agreed on the trust model — server-side validation is the right primary control here.

**Event link** — low severity, your call. The one nuance: this value lands in an `<a href>`, not an image, so a `javascript:`/`data:` scheme would execute on click rather than be inert, and we already run `safeHttpUrl` on a trusted-server URL for the community favicon (`community/[area]/+layout.svelte`). One-line change, non-http values just fall through to the plain-text branch. If the server validates the scheme it's only defense-in-depth. Not a blocker.

**Counter** — no need to remove it. The fragility isn't the cross-tab dependency, it's that `fetchAreaEvents` throws a 502 on failure and is awaited on every section, so one events hiccup blanks merchants/stats/activity/maintain. If it degrades to an empty list off the events tab (and only 502s when events is the active tab), you keep the badge and lose the blast radius. That also makes the places/open-issues counts you want safe to add later — each is another all-tabs fetch that shouldn't be able to take the page down.

via Claude Code 🤖

PR #1390, opened 2026-09-16
refactor(types): derive AreaEvent from the generated Event binding #1389

### LINKED ISSUE #1389: Derive AreaEvent from the generated Event binding instead of hand-writing it
### Background

#1351 added the area events section and hand-wrote `AreaEvent` in
`src/lib/types.ts`, with a comment explaining why:

> that struct has no `#[ts(export)]` annotation yet, so the type lives here
> instead of the generated `$types/btcmap-api/` bindings. When ts-rs is wired
> up on the API side, regenerate and import from `$types/btcmap-api/Event`
> instead.

Upstream added that annotation on 2026-09-10 (btcmap-api `e8c92804`), and the
binding is now committed here as `src/types/btcmap-api/Event.ts` — it arrived
with the bindings refresh in #1380.

So the TODO now looks actionable. It isn't, quite.

### Why a verbatim swap would be wrong

The generated type mis-describes three nullable fields as *absent-when-missing*
(`area_id?: bigint`, `ends_at?: string`, `cron_schedule?: string`), but the API
serializes them as `null` — see **teambtcmap/btcmap-api#113** and
`docs/rest/v4/events.md`, which documents "Integer or null" / "ISO 8601 datetime
or null".

Concretely: the fixture in `src/lib/areaSectionLoad.test.ts` mirrors upstream's
own documented example payload (`ends_at: null, cron_schedule: null`) and does
not typecheck against the generated `Event`. Adopting it as-is would mean either
weakening that fixture to something the API never sends, or having TypeScript
vouch for a shape that is wrong at runtime.

`area_id?: bigint` is wrong twice over — `JSON.parse` never produces `bigint`.

### Proposed change

Derive `AreaEvent` from the binding and correct only the mis-annotated fields,
so the six correct fields stay in lockstep with upstream:

```ts
import type { Event } from "$types/btcmap-api/Event";

export type AreaEvent = Omit<Event, "area_id" | "ends_at" | "cron_schedule"> & {
	area_id: number | null;
	ends_at: string | null;
	cron_schedule: string | null;
};
```

and replace the stale TODO with a note pointing at btcmap-api#113.

### Follow-up

Once btcmap-api#113 ships and `pnpm types:api` is re-run, this collapses to
`export type AreaEvent = Event` (or the alias is dropped entirely and call sites
import `Event`).

### Notes

No behaviour change: app code reads only `id`, `starts_at` and `ends_at` from
events — `area_id` and `cron_schedule` are carried but never consumed.

escapedcat:

Closes #1389. Reported upstream as teambtcmap/btcmap-api#113.

Stacked on #1380, which is where `src/types/btcmap-api/Event.ts` arrives — the
binding does not exist on `main`, so this targets that branch and should be
retargeted automatically once #1380 merges.

#1351 hand-wrote `AreaEvent` in `src/lib/types.ts` because btcmap-api's `Item`
struct had no `#[ts(export)]` annotation, leaving a TODO to import the generated
type once it existed. Upstream added the annotation on 2026-09-10 (`e8c92804`)
and the binding landed here with #1380's bindings refresh — so the TODO is now
reachable, but following it literally would be a regression.

Upstream marks three fields `#[ts(optional)]`, which in TypeScript means *the key
may be absent, and is never null*:

| Field | Generated | Actual wire | Corrected here |
|-------|-----------|-------------|----------------|
| `area_id` | `?: bigint` | `null` | `number \| null` |
| `ends_at` | `?: string` | `null` | `string \| null` |
| `cron_schedule` | `?: string` | `null` | `string \| null` |

None of them carries `#[serde(skip_serializing_if = "Option::is_none")]`, so
serde emits `null`. Upstream's own `docs/rest/v4/events.md` documents "Integer or
null" / "ISO 8601 datetime or null" and its example response contains literal
`"area_id": null, "ends_at": null`. `area_id` is additionally generated as
`bigint`, which `JSON.parse` never produces.

So `AreaEvent` now derives from the binding and corrects exactly those three
fields. The other six — `id`, `lat`, `lon`, `name`, `website`, `starts_at` —
come from upstream and will track it from now on.

The import is aliased (`Event as ApiEvent`) because this module already exports
its own v2 activity `Event`; without the alias the generated type shadows it and
`ActivityEvent` silently loses `element_id`/`created_at`. Same pattern as
`Place as ApiPlace` in `placementMode.ts`.

Once teambtcmap/btcmap-api#113 ships and `pnpm types:api` is re-run, this
collapses to `export type AreaEvent = ApiEvent`.

**Screenshots**

Not applicable — type-only change.

**Additional context**

No behaviour change: app code reads only `id`, `starts_at` and `ends_at` from
events; `area_id` and `cron_schedule` are carried but never consumed.

Verified locally: `tsc` clean, `svelte-check` 1270 files / 0 errors, 759 unit
tests passing. The fixture in `src/lib/areaSectionLoad.test.ts` mirrors
upstream's own documented example payload (`ends_at: null, cron_schedule: null`)
and does **not** typecheck against the raw binding — that fixture is the evidence
the correction is necessary rather than defensive.


escapedcat:
Closing without merging: the premise changed upstream. On 2026-09-17 btcmap-api resolved teambtcmap/btcmap-api#113 by changing the response instead of the type: `area_id` and `cron_schedule` are no longer returned (e8e42ed0), and `ends_at` is omitted rather than sent as `null` (3f6e8c00). The `| null` overrides here would now describe a payload the API no longer sends.

The generated `Event` binding is exact after a refresh, so #1403 replaces this with a plain `AreaEvent = ApiEvent` alias.

via Claude Code 🤖

Issue #1341, opened 2026-09-06
Add-location: review step before submit, captcha moves there


escapedcat:
Part of the add-location redesign (#1134) — the last unshipped storyboard screen besides the path fork.

Before submitting, show a "Here's what will be published" summary of the entered fields, with the captcha living on that step instead of gating the whole form.

## Why

- **Catches prefill junk** before it reaches the review queue — the reverse-geocoded address and generated opening hours get one explicit look.
- **The form stops being disabled while the captcha loads** — today every input waits on the captcha fetch; moving it to the review step lets people type immediately.
- Honest framing per the design deck: "this is exactly what editors see."

## Notes

- Lives inside the in-map panel (`AddPlaceFormPanel` hosting `AddLocationForm`) as a second view, not a separate page.
- Pairs naturally with the honest success track — likely one PR for both.

PR #1398, opened 2026-09-17
feat(add-location): step indicator in the panel header #1394

### LINKED ISSUE #1394: Add-location: step indicator in the panel header, back link at the top of review
**Is your feature request related to a problem? Please describe.**
The panel header says "Add Location" in both form steps, so nothing in the chrome changes when the form switches to review. The back action is an "Edit details" button at the very bottom, below the captcha — back and forward sit at opposite ends of a long scroll. Follow-up to #1341.

**Describe the solution you'd like**
The header gains "Step 1 of 2 · Details" / "Step 2 of 2 · Review" and a two-segment bar (`link` for reached, `input` for ahead). "Back to details" moves to the top of the review step; the bottom button goes. The ✕ keeps meaning "leave the form". The success screen shows no counter.

**Describe alternatives you've considered**
A numbered stepper in the body — it would scroll away with the content.

**Additional context**
From the add-location design review (F1).

escapedcat:

Closes #1394 (follow-up to #1341).

The form's progress lives in the panel header, and back sits next to it instead of at the bottom of a long scroll.

- `AddPlaceFormPanel.svelte`: the header shows "Step 1 of 2 · Details" / "Step 2 of 2 · Review" (`aria-live="polite"`) and a two-segment bar, `link` for reached and `input` for ahead (`aria-hidden`). Neither shows on the success screen. The ✕ is unchanged.
- `AddLocationForm.svelte`:
  - "Back to details" (chevron, link style; same `backToEdit` handler and `add_place_review_back` event) is now at the top of the review step, and the bottom "Edit details" button is removed.
  - The form's scroll margin goes from `scroll-mt-16` to `scroll-mt-24`. The taller header otherwise left the back link 14px under it when review opened. Deploy-preview QA caught this, and the e2e now asserts it.
- i18n, all 9 locales: added `addLocation.stepIndicator` (`{current}`, `{total}`, `{name}`), `addLocation.stepDetails`, `addLocation.stepReview`, `addLocation.reviewBackButton`; removed `addLocation.reviewEditButton`.
- e2e (`tests/add-location-review.spec.ts`): the round-trip test asserts the step label on both steps, the back link above the summary and clear of the sticky header, and that the old "Edit details" button is gone. Both assertions were pushed ahead of their fixes and failed first; the header overlap failed with `Received: -14`.

**Screenshots**

Deploy preview. Edit step at 1280×720:

Edit step with the step indicator, desktop

Review step at 390×844, light / dark:

Review step with back link, light
Review step with back link, dark

**Additional context**

One of four PRs implementing the add-location design review: #1397 (F3), #1398 (this one, F1), #1399 (F4), #1400 (F2). They merge in any order; with #1399, the success screen correctly shows no step label.

Unrelated, noticed during QA: the captcha label renders "Bot protection ((case-sensitive))". `forms.captchaCaseSensitive` already contains the parentheses and the template adds another pair. It's the same on `main`, so it's left out of this PR.


dadofsambonzuki:
All good, but I noticed on mobile there is no way obvious way to change the pin location (despite the text) as the map is not visible. Need to think how we best handle that, perhaps with a specific mobile action.

escapedcat:
> All good, but I noticed on mobile there is no way obvious way to change the pin location (despite the text) as the map is not visible. Need to think how we best handle that, perhaps with a specific mobile action.

In which situation is this needed? When the user filled out the form, and then, notices that the address is off? They can  still edit the address?

dadofsambonzuki:
>In which situation is this needed? When the user filled out the form, and then, notices that the address is off? They can still edit the address?

Yeah. Right now it's just that the text explicitly states it: `Add the business details below — you can still fine-tune the pin.` Which is only true on desktop.

The only way to get back to map screen on mobile is multiple swipes back.

So we need an 'alter pin location' flow on mobile or drop the misleading text. Preference is former.

escapedcat:
> So we need an 'alter pin location' flow on mobile or drop the misleading text. Preference is former.

That flow is in #1400 (F2). This PR (#1398) only adds the step header, which is why the preview here doesn't show it. #1400 also removes the "…you can still fine-tune the pin" text.

On mobile the form shows the pin as a chip with **Move pin**. Tapping it hides the form without discarding it and brings the map back with the crosshair and a "Move the pin" sheet. *Use this position* returns to the same form, with everything you typed still filled in.

Checked on the #1400 deploy preview at 390×844: filled in name, category, On-chain and email → Move pin → dragged the map → Use this position. All fields were kept, and the chip showed the new coordinates.

1. Form with the pin chip:

Form with the pin chip

2. Move pin: map, crosshair and sheet:

Move the pin sheet on mobile

3. Back on the form: fields kept, new position:

Back on the form with fields kept

Try it: https://deploy-preview-1400--btcmap.netlify.app/map?add=nav

via Claude Code 🤖

escapedcat:
Machine was ahead

dadofsambonzuki (approved):

PR #1418, opened 2026-09-19
SPIKE — do not merge: /communities/add on TanStack Form

escapedcat:
Throwaway comparison, to be closed: the same page as #1415, with its form state and validation wired through `@tanstack/svelte-form` 1.33.5 instead of the in-house helpers. The rules function, field components and a11y helpers are unchanged, so only the wiring differs. It's here to run `tests/add-community-validation.spec.ts` against it in CI.

Not for review or merge.


escapedcat:
Closing the spike: TanStack Form was chosen, and this page now lives in #1415 on the shared `$lib/ruleValidation` (#1412).

via Claude Code 🤖

PR #1419, opened 2026-09-19
SPIKE — do not merge: /communities/add on an in-house createValidation helper

escapedcat:
Throwaway comparison, to be closed: the same page as #1415, with its per-form validation wiring (errors state, re-check, submit gating, focus map) folded into a ~40-line runes helper, `src/lib/createValidation.svelte.ts`, with no dependency. The other variant is #1418 (TanStack Form). Here to run `tests/add-community-validation.spec.ts` against it in CI.

Not for review or merge.


escapedcat:
Closing the spike: TanStack Form was chosen over this in-house helper (#1418); the forms moved onto it in #1412–#1417 and #1421.

via Claude Code 🤖

